PR #4, opened 2019-11-04
POEM_001 for units updates

robfalck:
POEM 001

Units updates (includes some useful astrodynamics, suggestions of units from other users welcome).

### DIFF
--- a/POEM_001.md
+++ b/POEM_001.md
@@ -1,6 +1,6 @@
 POEM ID: 001  
 Title: Units update  
-authors: [robfalck] ([real name] -- optional)   
+authors: [robfalck]
 Competing POEMs: N/A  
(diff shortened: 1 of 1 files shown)


POEM_001.md
```diff
@@ -0,0 +1,42 @@
+POEM ID: 001  
+Title: Units update  
+authors: [robfalck] ([real name] -- optional)
```
JustinSGray: You can remove the [real name] part, if you don't want to include it

POEM_001.md
```diff
@@ -0,0 +1,42 @@
+POEM ID: 001  
+Title: Units update
```
JustinSGray: Suggest: "Units update for better astrodynamics support"

PR #7, opened 2019-11-11
POEM_003: addition of a pre_setup method to create communicators and objects

anilyil:


POEM_003.md
```diff
@@ -0,0 +1,111 @@
...
+Suggested solution
+==================
+
+We suggest adding a `pre_setup` method to the `Problem` class in OpenMDAO.
+This method will:
+
+1. Create the communicators,
+2. traverse down the model hierarchy and call the `pre_setup` method on all components, and
+3. traverse up the model hierarchy and call the `pre_configure` method on all components.
+
```
JustinSGray: This description currently matches what is done in the `setup` method. When combined with the existing `configure` method I think you have all the functionality you need.

POEM_003.md
```diff
@@ -0,0 +1,111 @@
...
+After `pre_setup` is called, the user will still be able to make new connections in the model, and define new input and outputs for any component.
```
JustinSGray: The existing `configure` method already traverse up the model hierarhchy. `setup` traverses down, assigning comms to all group and component instances, and then `configure` is called on the way back up. 

You are allowed to define new i/o for components and new connections for groups each ones `configure` implementation.

POEM_003.md
```diff
@@ -0,0 +1,111 @@
...
+# call pre_setup
+# all communicators and components are initialized with this method
+p.pre_setup()
+
+# add geometry design variables
+# this method also adds 'twist' as an output of the comp
+p.model.geometry.addDVGeoGlobal('twist')
+
+# get surface coordinates from solver
+surfaceCoords = p.model.solver.getSurfaceCoordinates()
```
JustinSGray: This is a core question that I want to understand better, before I can comment more fully on this POEM: 

In lines 76 and 79, you're calling methods on the component or group instances that you created above. How critical do you feel that it is that these methods be called on directly on the OpenMDAO derived sub-classes themselves, instead of on delegate objects that reside in those components. In other words, could you call those methods on generic python classes that get passed into the OpenMDAO component/groups? 

At my first reading these two calls seem like solver/geometry tool specific calls that are not necessarily general to all solvers or geometry tools.

JustinSGray:
Thanks for taking the time to work on this topic, which related to [this topic in the 2020 development roadmap](https://github.com/OpenMDAO/OpenMDAO/blob/master/ROADMAP.md#2-make-low-level-modifications-to-how-openmdao-handles-data-passing-in-key-situations-when-copying-can-be-avoided). 

Based on my initial reading of this POEM, I think you can already accomplish all of the requested functionality by using a combination of the `setup` and `configure` methods. 
See this [theory manual entry](http://openmdao.org/twodocs/versions/2.9.1/theory_manual/setup_stack.html?highlight=configure) and [this doc](http://openmdao.org/twodocs/versions/2.9.1/features/core_features/grouping_components/configure_method.html?highlight=configure) for more details

Can you have a look at these docs and at least adjust your POEM to use these method names. If you still feel that the methods don't provide the ability to handle needed functionality, can you elaborate more with a new test case?

POEM_003.md
anilyil: Currently, we cannot add new i/o and make new connections after the p.setup() call finishes. We want to be able to create new i/o, and make new connections outside the p.setup() call.

POEM_003.md
anilyil: These calls are not solver specific and they define an API that enables the use of a geometry tool with a solver. We can define other interfaces, but the root problem persists; we need to be able to take the initial aerodynamic problem surface definition and add it as an output to the geometry tool, which will be upstream of the aerodynamic solver.

anilyil:
> Thanks for taking the time to work on this topic, which related to [this topic in the 2020 development roadmap](https://github.com/OpenMDAO/OpenMDAO/blob/master/ROADMAP.md#2-make-low-level-modifications-to-how-openmdao-handles-data-passing-in-key-situations-when-copying-can-be-avoided).
> 
> Based on my initial reading of this POEM, I think you can already accomplish all of the requested functionality by using a combination of the `setup` and `configure` methods.
> See this [theory manual entry](http://openmdao.org/twodocs/versions/2.9.1/theory_manual/setup_stack.html?highlight=configure) and [this doc](http://openmdao.org/twodocs/versions/2.9.1/features/core_features/grouping_components/configure_method.html?highlight=configure) for more details
> 
> Can you have a look at these docs and at least adjust your POEM to use these method names. If you still feel that the methods don't provide the ability to handle needed functionality, can you elaborate more with a new test case?

Most of the functionality can be achieved with the current `setup` and `configure` methods; however, this approach requires the geometry and solver objects to be created under a single group. With the current implementation, we cannot create one geometry object on the problem level, and have multiple groups of aerodynamic solvers that take the surface definition from this single geometry tool. 

Adding another group on top of hierarchy temporarily solves the problem, but in general, this method requires at least writing a new group whereas all of these components could live on the problem level for a simpler implementation.

POEM_003.md
JustinSGray: ok, so in the case you're presenting, the user does all this in the run-script itself, after they call problem setup. 

You could currently work around this by defining a group sub-class in your run script where you put your setup and configure actions. Then set that group as the root of the problem. That would give you the desired functionality. However, you could potentially argue that its a bit clunky that way. 

So clearly the functionality already exists in OM, but it seems like this POEM is asking for more granular access to the setup stack then, to access the needed functionality without having to sub-class anything.

anilyil: > ok, so in the case you're presenting, the user does all this in the run-script itself, after they call problem setup.
> 
> You could currently work around this by defining a group sub-class in your run script where you put your setup and configure actions. Then set that group as the root of the problem. That would give you the desired functionality. However, you could potentially argue that its a bit clunky that way.
>
> So clearly the functionality already exists in OM, but it seems like this POEM is asking for more granular access to the setup stack then, to access the needed functionality without having to sub-class anything.

Exactly. We can get around it, but it creates additional levels that are not necessary and complicates the run scripts.

JustinSGray:
```Adding another group on top of hierarchy temporarily solves the problem, but in general, this method requires at least writing a new group whereas all of these components could live on the problem level for a simpler implementation.
```

That is not technically 100% correct. The problem is created with an empty group as the `model` attribute. Lets say for the sake of argument that you used `pre_setup` to accomplish your goals. You would still have the top level group with stuff in it. 

You could have also used the existing functionality and defined a new top-level group to your liking as a sub-class of Group (including your one geometry class and many aero instances, plus all indep-var-comps, etc). Then you could set that group to be the `model` attribute of the problem and you'd have the exact same final outcome as if you used pre_setup. 

Still, your point is valid. Our current API is forcing you to sub-class group to get things done. While this is a valid approach, its perhaps not the most elegant. We can discuss the consequences of giving a more granular access to the setup stack to enable your requested functionality. In fact, this was the original intent of the whole configure method itself. We wanted users to be able to change solver settings (as in OpenMDAO's nonlinear and linear solver) and set up case recorders after `setup` was called but before `final_setup` happened. 

So perhaps all that is needed here is a more deliberate and careful rethinking of how users can call the setup stack. I would still want simple cases to be handled by a single `setup` call. But it would be nice if we had a more advanced setup stack that could be used for parallel or even complex serial cases at the users discretion. 

The only hitch I really see is that this violated the general tenant of "only have one way to do things", but thats not really a hard and fast rule. More of a guideline for good API design. Still in that regard its worth asking: 

**Is it really worth adding a new more granular setup stack, simply to avoid sub-classing Group when necessary?**
I genuinely don't have a solid answer to this question myself, but it seems at the moment this is the core issue of this POEM. Thoughts?

anilyil:
> The only hitch I really see is that this violated the general tenant of "only have one way to do things", but thats not really a hard and fast rule. More of a guideline for good API design. 

Look at this as an alternate way of getting more functionality. All of the code written with OpenMDAO up to now calls `setup()` and that is enough. With the inclusion of `pre_setup()`, this does not change. During `setup()`, you an check internally if `pre_setup()` is already called and if it is not yet called, you can either call it internally or perform the object creation in `setup()` as it is done now, and if it is called, you can skip the object creation during `setup()`.

For run scripts that require this additional feature, `pre_setup()` will provide access to the objects before `setup()` is called. There is still one way to do things: It is to call `setup()` before you run your model. However, if you need more functionality, it is available.

> **Is it really worth adding a new more granular setup stack, simply to avoid sub-classing Group when necessary?**
> 
> I genuinely don't have a solid answer to this question myself, but it seems at the moment this is the core issue of this POEM. Thoughts?

Yes, this is the core issue this POEM tries to fix. As for the previous question on importance, we don't know yet. If doing things using the current functionality is simple enough and it is not limiting the functionality, then you can argue that the `pre_setup()` addition is not required.

However, as we are building more complex models and require more modularity, the code complexity can start becoming an issue. Furthermore, while the capabilities are there, it is going to be humans that are doing the model creation. In the future, we might see that the current approach increasing the entry barrier due to the extra steps required to set up a model. Again, this can be a non-issue, but it is too early to tell now.

JustinSGray:
Here is an alternative way to structure your setup script, using existing OpenMDAO functionality with `setup` and `configure` methods on the components. 

```
import openmdao.api as om 

from MACH_Aero import DVGeometryComp, ADflowComp
from TACS import TACSComp

# re-usable FSI group
class FSI(om.Group): 

    def setup(self): 

        # add solver component
        self.add_subsystem('aero_solver', ADflowComp(load_io=True))
        self.add_subsystem('struct_solver', TACSComp(displacement_io=True))

    def configure(self): 

        self.connect('aero_solver.loads', 'struc_solver.loads')
        self.connect('struct_solver.disp', 'aero_solver.disp')

        # NOTE: In reality, you probably have special load/disp transfer 
        #       classes that are involved here, but they are omitted for 
        #       clarity because they would not change the overall design 
        #       pattern

# non-reusable top level class
class Top(om.Group): 

    def setup(self): 

        # add geometry component
        self.add_subsystem('geometry', DVGeometryComp())

        self.add_subsystem('aero_struct', FSI())

    def configure(self): 

        # get surface coordinates from solver
        aero_coords = self.aero_solver.getSurfaceCoordinates()
        struct_coords = self.struct_solver.getSurfaceCoordinates()

        self.geometry.addPointSet(aero_coords, 'cfd_surface')
        self.geometry.addPointSet(struct_coords, 'struct_coords')
        # causes a call to add_output in the geometry component

        # connect surface output from geometry to the CFD
        self.connect('geometry.cfd_surface', 'aero_struct.cfd_surface')
        self.connect('geometry.struct_surface', 'aero_struct.struct_surface')

# initialize the OpenMDAO problem
p = om.Problem()

p.model = Top()

# call pre_setup
# all communicators and components are initialized with this method
p.setup()


# run
p.run_model()
```

anilyil:
> Here is an alternative way to structure your setup script, using existing OpenMDAO functionality with `setup` and `configure` methods on the components.
> 

I agree with this, and will give it a try. I am keeping the PR open so that we can keep having the discussion, and involve other people who might have more input on this.

fzahle:
Hi @anilyil and @JustinSGray 
We have faced the same challenge when setting up an OpenMDAO workflow around our FFD/CFD workflow. Both the geometry and the CFD codes work on the distributed mesh, therefore with each proc working on a local mesh size determined by the size of the COMM, and determined after  initialising the CFD code.

Our implementation seems similar to what @JustinSGray suggests, although I've not made use of the `configure` method yet. We define a `FlowGroup` that is nested inside a `ParallelGroup`, and we duplicate the FFD component across a multipoint setup, since the mesh data in the FFD and CFD comps then match in size 1:1 on a given sub-comm. The not so clean part of our setup is that we need to set the FFD and CFD codes' COMMs and initialise them in the top-level `FlowGroup`'s setup method. If we had a `pre_setup` method, I could move all operations on the FFD/CFD codes to their respective components, except for getting/setting the initial mesh that the FFD should operate on. So although we can work around it in the current setup stack, I agree with @anilyil that it could become a little more clean and modular with a pre_setup method. But maybe it is very specific to our code structure and CFD->FFD dependencies.

Below is some code to illustrate how a before/after workflow could look for us.

## Current setup

    class FlowGroup(Group):

        def __init__(self):
            super(FlowGroup, self).__init__()

            # do some stuff

        def setup(self):

            self.add_subsystem('FFDcomp', FFD())
            self.add_subsystem('cfdcomp', CFD())
            # and connect the output FFD mesh with the e3d comp's corresponding input
            self.connect('FFDcomp.mesh:x', 'cfdcomp.mesh:x')
            self.connect('FFDcomp.mesh:y', 'cfdcomp.mesh:y')
            self.connect('FFDcomp.mesh:z', 'cfdcomp.mesh:z')

           # I would really like that the CFD mesh readin and initialisation could be done
           # in the CFD comp but we have to do it here to initialise all internal data
           # in the CFD solver in order to setup the OpenMDAO I/O sizes
           self.FFDcomp.ffd.set_mpi_comm(self.comm)
           self.cfdcomp.cfd.set_mpi_comm(self.comm)
           self.cfdcomp.cfd.readin()
           self.cfdcomp.cfd.initilize()

           # extract local mesh
           x, y, z = self.cfdcomp.get_mesh()
           
            # set the initial mesh in FFD comp
            self.FFDcomp.set_mesh(x, y, z)


    class FFD(ExplicitComponent):

        def __init__(self, nCPs):
            super(FFD, self).__init__()

            self.ffd = FFDCode()
            self._nCPs = nCPs

        def set_mesh(self, x, y, z):

            self.x = x
            self.y = y
            self.z = z

        def setup(self):

            self.add_input('CPs', np.zeros(self._nCPs))
        
            self.add_output('mesh:x', np.zeros(self.x.shape))
            self.add_output('mesh:y', np.zeros(self.x.shape))
            self.add_output('mesh:z', np.zeros(self.x.shape))

            self.ffd.initialize(self.x, self.y, self.z)

    class CFD(ExplicitComponent):

        def __init__(self):
            super(CFD, self).__init__()

            self.cfd = CFDCode()

        def setup(self):

            # mesh input        
            self.add_input('mesh:x', np.zeros(self.cfd.x.shape))
            self.add_input('mesh:y', np.zeros(self.cfd.x.shape))
            self.add_input('mesh:z', np.zeros(self.cfd.x.shape))

            # and some outputs...

    # initialize the OpenMDAO problem
    p = om.Problem()

    # add indepvars for DVs and parameters

    pg = p.model.add_subsystem('pg', ParallelGroup())

    for case in range(ncases):
        pg.add_subsystem('case%02d' % case, FlowGroup(...))
        # connect DVs into FFD comp

    p.setup()


    # run
    p.run_model() 


## With pre-setup

    class FlowGroup(Group):

        def __init__(self):
            super(FlowGroup, self).__init__()

            # do some stuff

        def setup(self):

            self.add_subsystem('FFDcomp', FFD())
            self.add_subsystem('cfdcomp', CFD())
            # and connect the output FFD mesh with the e3d comp's corresponding input
            self.connect('FFDcomp.mesh:x', 'cfdcomp.mesh:x')
            self.connect('FFDcomp.mesh:y', 'cfdcomp.mesh:y')
            self.connect('FFDcomp.mesh:z', 'cfdcomp.mesh:z')

           # extract local mesh
           x, y, z = self.cfdcomp.get_mesh()
           
            # set the initial mesh in FFD comp
            self.FFDcomp.set_mesh(x, y, z)


    class FFD(ExplicitComponent):

        def __init__(self, nCPs):
            super(FFD, self).__init__()

            self.ffd_object = FFDCode()
            self._nCPs = nCPs

        def set_mesh(self, x, y, z):

            self.x = x
            self.y = y
            self.z = z

        def pre_setup(self):

           self.ffd.set_mpi_comm(self.comm)

        def setup(self):

            self.add_input('CPs', np.zeros(self._nCPs))
        
            self.add_output('mesh:x', np.zeros(self.x.shape))
            self.add_output('mesh:y', np.zeros(self.x.shape))
            self.add_output('mesh:z', np.zeros(self.x.shape))

            self.ffd.initialize(self.x, self.y, self.z)


    class CFD(ExplicitComponent):

        def __init__(self):
            super(CFD, self).__init__()

            self.cfd = CFDCode()

        def get_mesh(self):

            self.mesh = self.cfd.get_mesh()
            return self.mesh

        def pre_setup(self):

           self.cfd.set_mpi_comm(self.comm)
           self.cfd.readin()
           self.cfd.initilize()
           self.cfd.get_mesh()

        def setup(self):

            # mesh input        
            self.add_input('mesh:x', np.zeros(self.cfd.x.shape))
            self.add_input('mesh:y', np.zeros(self.cfd.x.shape))
            self.add_input('mesh:z', np.zeros(self.cfd.x.shape))

            # and some outputs...

    # initialize the OpenMDAO problem
    p = om.Problem()

    # add indepvars for DVs and parameters

    pg = p.model.add_subsystem('pg', ParallelGroup())

    for case in range(ncases):
        pg.add_subsystem('case%02d' % case, FlowGroup(...))
        # connect DVs into FFD comp

    p.setup()

    # run
    p.run_model()

JustinSGray:
Thanks for this input @fzahle. In your sample code, if at the component level you switched to `pre_setup` to `setup` and `setup` to `configure`, it seems like your code would align with the proposed API. 

Now, if we're agreed on the general form and timing of the API, the next question. Does anyone have better names for these methods now? `setup` and `configure` lack some descriptiveness. We could consider re-naming them with deprecation of the existing method names if better ones are available.

fzahle:
@JustinSGray ok, so you are proposing that `ExplicitComponent` can also define a `configure` method or what? And in this `configure` method, components can define their I/O?

As to the naming, it would be good if it indicated an obvious sequence in which the methods are called. I may be guilty of not having read the docs properly, but kept forgetting when what is called.

JustinSGray:
Yes, all components (including explicit and implicit) would have configure methods they can use to create I/O. However, as the example shows you could also create the I/O on components during the parent groups `configure`. 

So at the component level `setup` goes down the hierarchy (breadth first search). Then, next, `configure` gets called from the bottom up. 

so perhaps, for clarity `setup_top_down` and `setup_bottom_up` would perhaps be better. we could maybe shorten to `setup_down` and `setup_up`

fzahle:
ok, got it. however, I don't like that parent groups define I/O on behalf of components. then it quickly gets unclear what I/O a component has.

`setup_down` and `setup_up` sort of assumes users understand that OpenMDAO first traverses down then up. that's not immediately intuitive... something with `pre_`/`post_` gives more clear indication of order.

JustinSGray:
The `pre` step is really the hierarchy construction step. i.e. Its the top down the construction of the hierarchy (which does not exist in its entirety till after the full `pre` recursion is complete). The `post` step then just does a bottom up traversal of the newly created hierarchy. 

So what are we `pre` and `post` of? I don't think that `pre` is the right modifier there, since the function that is called then is part of the creation step. `post_hierarchy_creation` works, but is wordy. 

I am beginning to suspect that perfect names for these methods simply won't exist. At the moment I think `setup_hierarchy` and `post_setup_hierarchy`. This would be renaming existing methods as follows: `setup`-> `setup_hierarchy` and `configure`->`post_setup_hierarchy`

It does seem like we're in broad agreement about the basic functionality fo these methods, whatever they are called. Most users can continue to use `setup_hierarchy` as they currently do. However, if you have more complex needs that require your children to be in existence, then you can wait till `post_setup_hierarchy` to do your I/O creation and connection definition. 

If you want to have group level `post_setup_hierarchy` calling methods to create I/O on components you can. However, if you don't like that idea, then you're also welcome to use the `post_setup_hierarchy` at the component levels to do that work for you. Both styles of group/component design are supported.

robfalck:
After discussing this with the development team we'd like to maintain the same Group method names that we have now. (`setup` instead of `setup_hierarchy` and `configure` instead of `setup_IO`).  This keeps the setup stack simple for simple use cases, and we didn't want to break/deprecate existing code when our potential replacement names for setup/configure still aren't descriptive enough.  We'll have to address that problem via documentation.

Functionally, we've allowed `add_input/add_output` to be called during `configure`.  Now the only difference is that `add_subsystem` *must* be called during setup, and may not be called during `configure`.

Originally I wrote our implementation of this as the PR for POEM_009 but I just rejected it to merge it here.  Our approach is going to be the following:
- raise an exception if `add_subsystem` is called during configure.
- Add a new `promotes` method to allow IO to be promoted during configure.
- Better document the use case for both setup and configure.  Users can still use only setup for the simple cases, but use configure when necessary.

Does this satisfactorily address the issue @anilyil & @fzahle?

anilyil:
This works for us, @robfalck

POEM_003.md
```diff
@@ -0,0 +1,152 @@
+POEM ID: 003  
+Title: Adding a pre_setup method to create objects and communicators
```
JustinSGray: Need to update the title of this poem now. Suggest "Allowing addition if I/O during Configure"

anilyil: Makes sense. Just changed it

PR #31, opened 2020-03-13
POEM_015: Automatic IndepVarComp outputs for all unconnected inputs

JustinSGray:


ehariton:
This will be useful to reduce the total lines of code required to be written when designing large optimization problems.

bbrelje:
I think this is an important upgrade. IndepVarComp is one of the least intuitive things about the API for new users

gbarter:
We (NREL wind team) fully support this change.  IndepVarComp is not only a confusing concept to new users, it has dragged down our ability to write modular components that can easily mix-and-match with each other.

JustinSGray:
@gbarter there are a few backwards incompatibilities listed in this poem. I want to make sure you saw one in particular: 

```
If you have an any promoted, but unconnected inputs with different values/units then you must call `add_input` at the group level. 
Otherwise an error will be thrown during setup, because the it is not possible to infer the units/value for the auto_ivc output. 
```

Im think this one will actually impact some of your code, and I wanted to get your thoughts. It's a pretty minor addition, but it might need to be made in multiple places. Its great if you think the improved functionality is worth the headache, but I just wanted to make sure?

gbarter:
Hi Justin,

We are two thumbs up on this POEM.  We do our best to code with consistent units and not rely on the auto-conversion features of OpenMDAO, although we are certainly not perfect.  I am not concerned about this backwards incompatibility, but I very much appreciate you checking on that!

Cheers,
Garrett

robfalck:
I'd appreciate feedback on the following change to this POEM.

`Group.add_input` is problematic for two reasons.  First, it's not technically adding an input to the group.  It's disambiguating conflicting metadata from promoted inputs within, should a corresponding auto_ivc output need to be generated.  Second, it may lead the user to wonder why `Group.add_output` doesn't exist. While we could add `add_output` as syntactic sugar for promotion...

 The simpler solution seems to be to rename `add_input` to `set_input_defaults`.  This makes the functionality of the method more clear.  But is that a good thing in this case?

johnjasa:
I think the name change is a good idea, or at the very least not calling it `Group.add_input`. Because `Group.add_input` would not produce any inputs without components having their own `add_input` calls, right? That would be misleading -- naively I'd probably try to add an input only at the group level and not at the component level, then try to use those inputs without components. So a different name is necessary in my understanding.

sanbales:
> 
> 
> I'd appreciate feedback on the following change to this POEM.
> 
> `Group.add_input` is problematic for two reasons. First, it's not technically adding an input to the group. It's disambiguating conflicting metadata from promoted inputs within, should a corresponding auto_ivc output need to be generated. Second, it may lead the user to wonder why `Group.add_output` doesn't exist. While we could add `add_output` as syntactic sugar for promotion...
> 
> The simpler solution seems to be to rename `add_input` to `set_input_defaults`. This makes the functionality of the method more clear. But is that a good thing in this case?

How about `Group.define_input` and `Group.promote_output`?

robfalck:
`Group.promotes` already exists for the purposes of promoting outputs.  In my humble opinion, `define_input` seems a bit too much like `add_input`.

sanbales:
> 
> 
> `Group.promotes` already exists for the purposes of promoting outputs. In my humble opinion, `define_input` seems a bit too much like `add_input`.

Sorry Rob, I don't know why I had only seen the first paragraph of your comment and not the last. I I do like `set_input_defaults` over `add_input` or `define_input`.

PR #37, opened 2020-03-25
POEM_019: Random Vectors in Directional Derivatives

kejacobson:


Kenneth-T-Moore:
I am going to start working on this, but wanted to ask you. Is there really any value in keeping the current all ones implementation as an option?  It seems like the random step would always be better.  Also, I don't think we need the seeded random, as we can always set the seed through numpy (which we do in tests where we need reproducible answers.)

kejacobson:
Good point about seeding. I can't think of a case where the uniform vector is better than a random vector.

Kenneth-T-Moore:
I've got a pull request up with this POEM implemented. Can you check it out and see if satisfies your goals for this POEM?  @kejacobson 

https://github.com/OpenMDAO/OpenMDAO/pull/1314

kejacobson:
@Kenneth-T-Moore,  in my testing I've found a component where I get match to machine zero without directional derivatives, but the directional derivative dot product test says they don't match

Directional Check:
```
-------------------------------------                                            
Component: MELD_disp_xfer 'disp_xfer'                                            
-------------------------------------                                            
  disp_xfer: 'u_a' wrt (d)'u_s'                                                  
    Forward Magnitude : 2.066053e+00                                             
         Fd Magnitude : 2.066053e+00 (cs:None)                                   
    Absolute Error (Jfor  - Jfd) : 2.215589e-15                                  
    Absolute Error (Jrev Dot Product Test) : -1.215908e+01                       
                                                                                 
    Relative Error (Jfor  - Jfd) : 1.072378e-15 
.
.
.
```

Without Directional (compact print):
```
-------------------------------------                                            
Component: MELD_disp_xfer 'disp_xfer'                                            
-------------------------------------                                            
'<output>' wrt '<variable>' | fwd mag.   | rev mag.   | check mag. | a(fwd-chk) | a(rev-chk) | a(fwd-rev) | r(fwd-chk) | r(rev-chk) | r(fwd-rev)
------------------------------------------------------------------------------------------------------------------------------------------------
                                                                                 
'u_a'      wrt 'u_s'        | 3.4374e+00 | 3.4374e+00 | 3.4374e+00 | 3.6078e-15 | 3.6307e-15 | 2.0029e-16 | 1.0496e-15 | 1.0562e-15 | 5.8270e-17
.
.
.
```

I'm trying to figure out why it is happening.

Kenneth-T-Moore:
@kejacobson Perhaps if you can send me a model, i can take a look as well.

PR #51, opened 2020-05-06
POEM 22:Shape inputs/outputs by connection or copy from another component variable

joanibal:
See POEM_22.md for details
(resubmitted the PR because I accidentally made it a draft PR)

### DIFF
--- a/POEM_22.md
+++ b/POEM_22.md
@@ -154,7 +154,7 @@ class Interpolate(om.ExplicitComponent):
              shape: tuple
                  the shape of the variable
             """
-            shape = var_meta['u_coarse'].shape*2
+            shape = var_meta['u_coarse'].size*2
 
             return (shape)
 
@@ -210,7 +210,7 @@ class OuterProduct(om.ExplicitComponent):
 
 
         def get_mat_shape(var_meta):
-            return (var_meta['vec_1'], var_meta['vec_2'])
+            return (var_meta['vec_1'].size, var_meta['vec_2'].size)
 
         # the shape of this output is related by the function to the shape of the variables defined
         # in shape_depends on 
--- a/POEM_22.md
+++ b/POEM_22.md
@@ -1,0 +1,259 @@
+POEM ID:   022  
+Title:  Shape inputs/outputs by connection or function.   
+Authors: joanibal (Josh Anibal)   
+Competing POEMs: N/A  
+Related POEMs: N/A
+Associated implementation PR: None (yet)
+
+Status:
+
+- [x] Active
+- [ ] Requesting decision
+- [ ] Accepted
+- [ ] Rejected
+- [ ] Integrated
+
+
+# Motivation
+
+Modularity is a critical aspect of OpenMDAO since it is designed to facilitate the use of discipline based components in many different multi-disciplinary models. 
+In order to be modular often a component is required to have inputs and outputs of variable shape.
+For example, components of a PDE based analysis must work with variables that have a shape defined by a mesh file.
+
+
+Passing shape as an option 
+--------------------------
+One method of making components modular is to pass an additional variable as an option which specifies the shape of the input vectors. 
+To utilize this method, at the time the component is created the shape variable argument must be passed. 
+This solution has worked well for many applications and is generally the way to do things in OpenMDAO as shown by the many modular components included with OpenMDAO that use this method (DotProductComp for example). 
+
+However this method does not work if the inputs and outputs shapes are unknown when the component is created. 
+Consider the case where during the setup a mesh is read to create the inputs and outputs of a component.
+This could not happen during the initialization of the object if the method used to read the mesh requires the component's MPI comm, which is typically the case in practice.
+In this case, the shape of the mesh can not be passed as an argument to other subsystems as they are create because the  they are created before any of the system setup methods are run.
+
+
+Builders
+--------
+One method to get around this issue is to use a "builder" (see the MPHYS project).
+The builder object can read in the mesh and pass shape data to each of the components as they are created. 
+Builders can be passed as an arguments to other builders to share shape information between subsytems (in MPHYS this is used by the transfer components)
+
+However, this method greatly increases the complexity of the code and reduces modularity. 
+For each new model in which the components are used, a new builder must be made or an existing builder must be generalized (increasing complexity). 
+Additionally, Now instead of writing components, users must write custom classes which create components. 
+This is a different design pattern than what is used for low fidelity simulations and is likely to cause confusion for OpenMDAO users when they first try to incorporate a high fidelity model in OpenMDAO.
+
+
+Add variables during configure
+------------------------------
+A third method that exists is to add add the inputs and outputs of a unknown shape to the components during the configure method of a group.
+Because the configure method is called after the setup, the inputs and outputs can be added with correct shape. 
+
+The major draw back of this method is that the information about these inputs and outputs exists only at the group level, which sacrifices a great deal of modularity and clarity. 
+Now the components do not contain enough information to be used in a model by themselves.
+The information about their inputs and output must be added to the model configuration or they must be used in an existing group which supplies the remaining information in its configure.
+Because the inputs and outputs must be restated for every group or model the component is used in, modification to the inputs/outputs, such as adding a new variables or changing variable names, units, and or description must be be repeated for every group or model the component is used in. 
+An Additional source of frustration is that the description of the input/output included in their definition now exists outside of the component definition. 
+
+
+<!-- The final method I'm a aware of is to 
+Move code to read mesh file into initialize 
+
+Finally, 
+need mpi communicators 
+- can't create input/outputs in init
+- breaks trend of using the initialize method to declare options
+- shape information stored using attr or method which exist outside the API 
+- mesh readers must use the same attr/method names to be swapable 
+<!-- - all information is saved becuase it not known yet what information is needed  -->
+
+# Description
+
+The solution I propose to create modular components of variable shape is to allow inputs and outputs to get their shape information from their connections or 
+by a function. 
+
+shape by connection
+---------
+When two variables are connected their shapes must match. 
+Therefore, connecting a variable with a known shape to one of an unknown shape can be used to specify the unknown shape. 
+By adding a keyword to the input or output definition we can specify that the shape of the input or output should be determined in this way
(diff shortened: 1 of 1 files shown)


anilyil:
I like the idea, but I think we should discuss the use cases a bit more. Also, I have a few questions and items I want to bring up:

1. The biggest issue I can think of is the hierarchy of these connections. How is this determined? Can a component upstream have an output with a size dependency to a variable that is coming from a component downstream? Obviously you should not have a circular dependency. Would OpenMDAO automatically determine the order of connections to be made to make sure everything is defined, or would it be determined based on the order you specify the connections?

2. This approach will not work for the common example case of "geometry component upstream of CFD solver". For this, on top of the size of the arrays, we also need the node coordinates to initialize these values. If we don't provide the initial values to the geometry component, you cannot connect these to the CFD nodes. For this particular problem, you either need to use the configure level to communicate these coordinates, initialize the CFD before the geometry setup is finished.

3. How would this work for parallel connections? I guess for parallel outputs, this is fine since the output shape is all you need to figure out the vector. But then, how do you handle the case where a parallel input uses this feature? You need to specify the src_indices to make a fully defined connection.

joanibal:
Thanks for the feedback!

1. The shape of shape differed variables are resolved after all connections are made. The connection just adds information about the dependency of shape_by_connection variables. The information on dependencies can be encoded in the form of a tree that traversed from bottom to top. The shape dependencies are resolved in an order that isn't related to the model hierarchy. So, yes shape information could be passed "upstream".

2. That is true, but I would argue that you need a sperate mesh component. If you added a mesh component upstream of the geometry component then you could set up your model like this

```

              Shape DVs
+---------+   +
|         |   |  X_0
|   Mesh  +-----+
|         |   | |
+---------+   v v
           +---------+
           |         |
           |   Geo   +-----+ X
           |         |     |
           +---------+     v
                      +---------+
                      |         |
                      |   CFD   |
                      |         |
                      +---------+
```
With a geometry component something like this 
```python
class Geo(om.ExplicitComponent):
    """
    Calculates the sum of the elements of the input vector x
    """
    def setup(self):
        # the shape of the input is not known when the 
        # component is defined

        self.add_input('X_0', shape_by_connection=True)
        self.add_input('shape_DVs', shape=10) 
        
        def my_shape_func(var_meta):
             return (var_meta['X_0'].shape)

        self.add_output('X', shape_function=my_shape_func, shape_depends_on=['X'])
 
    def compute(inputs, outputs):
         # transform the original mesh coordinates 
         # according to the set of DVS to get the new mesh coordinates         
    .
    .
    .
```
I think this is a clearer model as it shows that the output mesh is a transformed version of the initial mesh which is an input. 

If you wanted the initial points only to be embedded once you could use an if statement with a call counter or some other method to check if it was already done (i have another POEM idea for an API method that is run once before the model starts calculating outputs to allocate memory and compute other reused values)

3.  I can't say I have fully thought through distributed variable-size inputs, but here are my thoughts about how src_indices could be resolved. 
- even distribute the elements of an array by default 
- accept a function that is used to determine the src_indices from an input_shape
I'm not as familiar with distributed components, so I was hoping the Dev team would have more thoughts about how this could be done. 
Maybe we could introduce a way to have a distributed output variable which must be connected to disturbed input with the same src_indices. Then the information about what portion of the output live on what proc doesn't have to be lost when all parts of the output are brought together at the end of the compute.

JustinSGray:
We've started to discuss this POEM within the dev team. It would help us a lot to get a clear set of use cases here, one for simple serial stuff. One for distributed, and one for distributed with src indices. 

Can you spend some time to think through the use cases? At least laying out what needs to be connected to what, perhaps in the context of mphys. That would give us the best hope of getting the API right. 

Right now it sounds like the primary issues are: 

1) What do you do with src_indices when the size of your inputs isn't known till late in the setup process. Currently, src-indices can be specified during the add_input call, or as part of the group connect/promote calls 

2) is there anything specific we need to do differently in the case of distributed outputs? 

If you can come up with some good example cases that we could use to turn into tests, that will help a lot.

joanibal:
I'd be happy to create this examples/tests cases for you. I'm in the middle of switching over to python 3, so it may take longer than normal. 
By use cases do you mean complete openMDAO components which can be run or something like the examples shown in the POEM?

As for the issues you raised..

1) I can't think of a situation in which the src_indices are known during the add_input, but the size of the input itself is not known. If you knew the largest src_index than you'd know the size of the input. For that reason I don't think it makes sense to allow an input to have a size which is given by a connection/function with src_indices which are given explicitly. Two ways I think we could provide methods for implicitly defining the src_indices are  
- through an additional function called after the size of the input is resolved
- Define a new type of output which has src_indices. This type of output must be connected to an another distrusted input. If that distrusted input is sized by connection then it can also gets it's src_indices from the connection too. This method passes information about src_indices through the model.

2) Do you mean the output of a normal distributed component or an output with src_indices like I talked about in 1) and in the answer to Anil?

JustinSGray:
By use cases I do mean sample input files with components and groups and connections. Mock it up, without any actual working implementation. Keep it simple to toy components that represent the I/O structure you'd expect to see from CFD. 

if this isn't making sense, we can discuss in more detail. @anilyil can also help a bit, because he's had to do this for me a few times. 

Making up whatever cases you think would be needed would be best.

joanibal:
@JustinSGray  and I talked offline about this proposal. The following comment is in response to Justin's request made offline to show how this feature would be used in MPhys, a high-fidelity simulation coupling framework.

I have uploaded how I would expect Mphys to change on my fork under the branch PEOM_22, which can be diffed with the master to highlight changes. 

# General Takeaways

In general, with this feature we can remove many redundant specifications of sizes and src_indices seen throughout the code. This can be seen in the specific example of the Geo_Disp component. The code using the features of this POEM is shown on the right (same for all pictures). 
image

By using the connections to infer size information, we now longer have to give the Geo_Disp component size information when it is created. This makes it easy for us to move the component outside of the ADflow group to the Multipoint group where it can be used with other solvers. 
image


However, within Mphys it is not always possible to create a completely modular components. This is because the aspects of the solvers which MPhys wraps were not originally programed in a modular way themselves but instead all implemented in one solver object. Thus, within a discipline group one solver object is created by a builder and shared among the discipline subcomponents. As a consequence we can not fully do away with the builder when a solver object is shared between components even when size information is not shared through the builder.

Although for cases like the load and displacement transfer components where creating a solver object for each component isn't particularly expensive we can now choose to remove the builder since it is not needed to distribute size information.  

Removing the need to use a builder from the Meld wrapper makes it easier to use Meld with other discipline components that do not use a builder. Before, to create the load and displacement transfer components one needed to give two other discipline builders to the Meld builder to provide size information. By allowing sizing information in a more natural way we can avoid this mess altogether. 
image

anilyil:
I am all for removing the repeated code to handle the src_indices stuff, I really like the examples here if it ends up working out. 

I dont agree with "Although for cases like the load and displacement transfer components where creating a solver object for each component isn't particularly expensive we can now choose to remove the builder since it is not needed to distribute size information."

If you remove the builder and create separate objects, on top of the memory penalty, you will incur the setup cost of these classes twice. I think for now it might be okay, but as we want to transition towards reinitializing these transfer schemes at every function evaluation, this cost will be significant.

I think we can come up with a smarter approach and avoid the specialization that is in the transfer builders. @joanibal I really liked your suggested approach towards generalized transfer schemes and components. I think @kejacobson also had a similar idea in mind. I suggest that direction would be better rather than removing the builders for transfer schemes. We can have this discussion using an issue in the MPhys repo.

joanibal:
I also wanted to provide you with the simplest possible serial case too.

# shape_by_connection
```python
import openmdao.api as om
import numpy as np

class Summer(om.ExplicitComponent):
    """
    Adds all the elements of a vector
    """
    def setup(self):
        self.add_input('x', shape_by_connection = True)

        self.add_output('sum', val=0.0)


    def compute(self, inputs, outputs):
        x = inputs['x']

        outputs['sum'] = np.sum(x)


if __name__ == "__main__":

    model = om.Group()
    ivc = om.IndepVarComp()
    ivc.add_output('x', np.ones(10))
    model.add_subsystem('des_vars', ivc)
    model.add_subsystem('summer', Summer())

    model.connect('des_vars.x', 'summer.x')

    prob = om.Problem(model)
    prob.setup()
    prob.run_model()
    print(prob['summer.sum'])
```

# shape by function / determine_size

```python
import openmdao.api as om
import numpy as np

class OuterProd(om.ExplicitComponent):
    """
    Calculates the outer product of two vectors. 
    """
    def setup(self):
        self.add_input('vec1', shape_by_connection = True)
        self.add_input('vec2', shape_by_connection = True)

        self.add_output('mat', determine_shape = True)

    def determine_shape(self, var_meta_data):
        self.set_variable_shape(['mat'], (var_meta_data['vec1'].size, var_meta_data['vec2'].size))

        # or perhaps something like

        var_meta_data['mat'].shape = (var_meta_data['vec1'].size, var_meta_data['vec2'].size)
        

    def compute(self, inputs, outputs):
        outputs['mat'] = np.outer(inputs['vec1'], inputs['vec2'])


if __name__ == "__main__":

    model = om.Group()
    ivc = om.IndepVarComp()
    ivc.add_output('x', np.ones(10))
    ivc.add_output('y', np.ones(3)*3)
    
    model.add_subsystem('des_vars', ivc)
    model.add_subsystem('outer', OuterProd())

    model.connect('des_vars.x', 'outer.vec1')
    model.connect('des_vars.y', 'outer.vec2')

    prob = om.Problem(model)
    prob.setup()
    prob.run_model()
    print(prob['outer.mat'])
```

joanibal:
@anilyil, I agree that using two solver objects is more expensive than one for the transfer schemes. However, I think that it is worth it. Lets talk more on the MPhys repo.

robfalck:
We're most likely going to be implementing the scaled back capability with `shape_by_conn` and `copy_shape` being options for inputs and outputs.  We'll need to revise this POEM to reflect that API.

This implementation isn't without its drawbacks:
1. The newly added `get_io_metadata`method, which is used for interrogating underlying systems for their inputs and outputs during configure, won't be able to return the shape of variables which use these features, since they won't be known at the time configure is run.
2. How do we declare sparse partials for variables when the shapes aren't known?

I think we're willing to include this implementation with this drawbacks, and possibly revamp them in the future to address these shortcomings.

naylor-b:
Regarding 2) above, we've added a function hook called `setup_partials` that doesn't get called until all shapes are known, so that problem is solved.

PR #55, opened 2020-06-01
POEM_024: Calculating ExecComp Jacobian with symbolic derivatives

onodip:


JustinSGray:
We actually used to use sympy in the execcomp a long time ago (I think way back in the early days of V1). I can't remember why we switched to CS though. I have a vauge recollection that it was actually a bit slower than just using CS. It might also have been harder to support corner cases like numpy methods. 

I don't fundamentally object to this idea, though Id want to see some evidence that the current CS method is actually a bottleneck before we added the complexity of rolling our own differentiation tool (even just using sympy is more complex than CS...) 

What was your motivation here? Do you have a case where its slowing you down a lot?

onodip:
My expectation was, that it will be faster, where the partial derivatives are much simpler than the function itself. And either with huge array inputs or in problems with many iterations, where the setup time gets negligible. I would't say though, that the ExecComp with CS was a real bottleneck anywhere. 

I will try to write some benchmark cases later.

robfalck:
@onodip Have you had a chance to benchmark this against complex-step?

onodip:
@robfalck not really, I only did some basic profiling. It wins in some cases (like the Sellar problem), but it spends much more time in `setup` and it is somewhat faster in `compute_partials`. So for simple models with fast convergence CS is better, I will also try it on something bigger. Later I will add some results.

PR #61, opened 2020-07-22
POEM_028: check_partials input warnings

ehariton:


Kenneth-T-Moore:
Any suggestions for a fix? In the past (with CADRE), I would check my partials after setting all of a component's inputs with random values.  This worked well in some cases, but other times I had to individually scale or shift ranges on each variable so that it would have a more typical value.

In addition to '0' and '1' being problematic, I have also found these:
- an array with all values the same (think mesh coordinates), creating a totally degenerate case
- using the same input value for multiple inputs causing subtractive cancellation
- default input value being way out-of-scale (1.0 for Young's Modulus in SI)

robfalck:
We could provide an option to randomly perturb the input values by some relative value, but as Ken said it won't resolve all cases - in particular, the "way out of scale" issue.

ehariton:
An option to randomize the inputs would be fine by me. However I think by default it should be off. I think this would require running `check_partials()` multiple times and then having the output information be a summary of all the cases run.

robfalck:
I don't think it would necessarily require multiple passes, although doing so would be more reliable.  Even just random perturbations of the variables should knock the default values off of 1.0 and significantly reduce the possibility of subtractive cancellation issues.

ehariton:
It would be useful for the user to know what input values caused the check to fail in the case of random perturbed inputs.

ehariton:
Based on what @robfalck suggested it sounds like a possible solution is to change the default behavior of `check_partials()` to perturb ALL the given input values and skip issuing a warning to the user. 

I looked through what IPOPT is doing and their documentation says "each component of the user-provided starting point is perturbed one of the other." Going to assume that's a typo for "one after the other". That type of implementation would be insufficient for our needs.

robfalck:
So the IPOPT explanation can be a bit confusing.  They are perturbing each input, one after the other...that's how finite differencing is performed.  But they also slightly perturb the nominal starting point of those perturbations (all of the design variables are initially perturbed in case they're on some "special" value).

If we were to do this, we'd want it to be optional.  Users should be able to check the partials at a given design point.

ehariton:
It sounds like these are the two options we have come up with:
A) Perturb =True by default. When Preturb = False Warnings on 0,1, or Equal inputs. 
B) Perturb = False by default and Warnings on 0,1, or Equal inputs. Perturb = True turns off warnings.  

I prefer option A. I feel like new users would have a better chance of successfully creating derivatives if there is a little magic in the background.

JustinSGray:
"Magic" is a dirty word for our development team. Things that are magic genereally lead to confusion. That being said, IMO default perturbation is ok and probably the safer option for most people. As long as they have the option to turn it off, Im good with it. 

However, I think the 0/1 warnings should be separate from the purtubation stuff. Any purturbation is not sure to solve the issue because we have no idea  how much to move it. If you think a warning is warranted, then you should get it depending on where the baseline point is (unless you set a flag to silence it). 

So purturbations and warnings should be separate arguments, I think

ehariton:
I updated the POEM based on the above suggestions.

robfalck:
I'd like to suggest that `nudge` be called `rel_initial_perturb` and it be a tuple of two floating point values (min, max) that is the relative initial perturbation of the input as seen by the optimizer, and it have a default value of (1.0E-6, 1.0E-3).  The perturbation of any given variable would be a random number between those two limits.  Setting them to `None` or (0, 0) would disable the initial perturbation.

robfalck:
After discussion, the OpenMDAO developer team decided that this behavior is best left to the developer when devising tests for their components.
The possibility of perturbing inputs in an invalid direction, and the possibility of confusing users who expect a certain partials at a given set of inputs makes for a lot of corner cases.

PR #62, opened 2020-07-27
POEM_029 - Proposal for a variable metadata retrieval function.

naylor-b:


JustinSGray:
@bbrelji @johnjasa @anilyil @kejacobsonId appreciate it if you took a look at this one. Its an api proposal for accessing i/o of componts AND groups inside the configure stack. 

Note: currently you can use list_inputs and list_outputs, but only on components. This POEM will alleviate that restriction, but we're proposing a slightly modified API

JustinSGray:
@bbrelji @johnjasa @anilyil @kejacobsonId appreciate it if you took a look at this one. Its an api proposal for accessing i/o of componts AND groups inside the configure stack. 

Note: currently you can use list_inputs and list_outputs, but only on components. This POEM will alleviate that restriction, but we're proposing a slightly modified API 

@naylor-b can you add some more detail to the motivation section to clarify the difference in defaults between list method and the get_metadata method under MPI situations?

anilyil:
Thanks for letting us know Justin, I will take a look as soon as when I can find the time.

@bbrelje @kejacobson, your usernames were misspelled, tagging you here so you get notified.

JustinSGray:
thanks @anilyil ... thats what I get for commenting while tired :)

johnjasa:
I like this idea, but I've never personally implemented `configure()` in any model, so take my opinion with a grain of salt. `get_io_metadata()` sounds well thought-out. I appreciate the inclusion of discrete variables. No improvements come to my mind, at the moment.

bbrelje:
I am already kind of doing something similar in OpenConcept. https://github.com/bbrelje/openconcept/blob/947bafc468e728b6bedd87b07c866a28f5eeeacc/openconcept/analysis/trajectories.py#L51

Maybe this breaks in parallel or something but for the time being it's my hack. I would support an "official" way of getting variable metadata. Because of the way _setup_var_data() is written it can be run multiple times without apparent side effects (it creates all new data each time). So implementing this poem should be easy and non breaking

PR #67, opened 2020-08-14
Removal of backwards incompatibility from POEM_029

robfalck:
The list_input and list_output function return values will remain as lists, rather than being changed to dictionaries.
While simpler, at this time this change is too significant, required potentially major changes to libraries which depend upon OpenMDAO.
In addition, changing these methods would be in conflict with the functionality of list_inputs/list_outputs in the CaseReader.

### DIFF
--- a/POEM_029.md
+++ b/POEM_029.md
@@ -19,8 +19,8 @@ Motivation
 
 As a result of changes to the OpenMDAO setup stack, it's now possible to access variable metadata
 from sub-groups and sub-components during `configure`.  The only functions that provide access
-to some of that metadata currently are `list_inputs` and `list_outputs`, which return a dict of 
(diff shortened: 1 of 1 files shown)


POEM_029.md
```diff
@@ -19,8 +19,8 @@ Motivation
 
 As a result of changes to the OpenMDAO setup stack, it's now possible to access variable metadata
 from sub-groups and sub-components during `configure`.  The only functions that provide access
-to some of that metadata currently are `list_inputs` and `list_outputs`, which return a dict of 
-metadata keyed by variable name and prints that information to an output stream.  However,
+to some of that metadata currently are `list_inputs` and `list_outputs`, which return a list of
+tuples of variable names and associated metadata and prints that information to an output stream.  However,
```
naylor-b, line 23: prints --> print

PR #68, opened 2020-08-18
Improvements to Atiken relaxation 

JustinSGray:
Add some extra user accessible options, and implement the algorithm in both nonlinear and linear solvers.


joanibal:
Why is the recommended starting value 0.4?

JustinSGray:
Wasn't that what u suggested it was?

joanibal:
Oh
0.4 is what has worked for the few problems for which I've used it, but I didn't mean to suggest that 0.4 should be the default for all problems. For example, for the Sellar problem the best initial factor is very close to 1.0.

Although, a lower initial factor can converge faster in some cases, it isn't always faster. So I don't think it makes sense to break backwards compatibility to change the default.

PR #73, opened 2020-09-08
Poem 022

naylor-b:
Make a few edits to Josh's original POEM and switched the reference to the implementation PR.

### DIFF
--- a/POEM_022.md
+++ b/POEM_022.md
@@ -7,8 +7,8 @@ Associated implementation PR: None (yet)
 
 Status:
 
-- [x] Active
-- [ ] Requesting decision
+- [ ] Active
+- [x] Requesting decision
 - [ ] Accepted
 - [ ] Rejected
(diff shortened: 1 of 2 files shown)


robfalck:
@joanibal Any chance you can experiment with this and see if our implementation meets your needs?

joanibal:
@robfalck Sorry, not yet. I've been preoccupied with teaching assistant responsibilities, but I'll take a look tonight. 
Thanks for all the work you two have done to make this happen.

joanibal:
I looked at the PR and I really like the implantation. It is written in a way that will make it easy to expand the feature in the future if needed. I have a few questions that I will leave in the PR on the openMDAO repo. 

I ran it on the case were I was using the prototype before and it worked just fine. My use case is covered by the tests I wrote for the prototype so I appreciate that you included them in the final design.

PR #78, opened 2020-10-05
added POEM 035

robfalck:


POEM_035/case3.py
```diff
@@ -0,0 +1,38 @@
...
+                         y={'shape': (4,)},
+                         a={'shape': (4,)})
+
+        p.model.add_subsystem('g1', g1, promotes_inputs=['a', 'b'])
+        g1.add_subsystem('c1', c1, promotes_inputs=[('a0', 'a'), 'b'], promotes_outputs=['y'])
+
+        g2.add_subsystem('c2', c2, promotes_inputs=['a', 'y'])
+        g1.add_subsystem('g2', g2)
+
+        g1.promotes('g2', inputs=['a'], src_indices=[0, 0, 0, 0], src_shape=(4,))
```
naylor-b, line 35: So promoting with src_indices=[0,0,0,0] at the g1 level results in promoted variable 'a' at g1 level having two different shapes, (1,) and (4,).  Shouldn't we promote using src_indices of [0,0,0,0] at the g2 level instead?  Then we could promote 'a' from g2 to g1 using src_indices=[0] making the shape of 'a' consistent at g1 level?

POEM_035.md
```diff
@@ -0,0 +1,44 @@
...
+![Current behavior for sizing a promoted input](POEM_035/current_behavior.png)
+
+Instead, this POEM proposes making src_indices apply to each level of promotion.
+During setup, OpenMDAO can then traverse the chain of connections and resolve which elements of the ultimate source end up where in the ultimate target.
+
+### Proposals
+
+1. `src_shape` is added as an argument to `promotes` and `add_subsystem` (when inputs are promoted), which specifies the apparent shape of the promoted input (if connecting to it) or the related AutoIVC output (if left unconnected).
+
+2. If `src_indices` are specified, then `src_shape` must be specified to remove any ambiguity, and `src_indices` are relative to that `src_shape`.
```
JustinSGray, line 38: This is backwards incompatible. Currently no user has specified src_shape. Did you mean it to be this way? 

Will we break old models?

POEM_035.md
```diff
@@ -0,0 +1,44 @@
...
+A user may unwittingly try to connect to that with `src_indices` and encounter an error.
+
+![Current behavior for sizing a promoted input](POEM_035/current_behavior.png)
+
+Instead, this POEM proposes making src_indices apply to each level of promotion.
+During setup, OpenMDAO can then traverse the chain of connections and resolve which elements of the ultimate source end up where in the ultimate target.
+
+### Proposals
+
+1. `src_shape` is added as an argument to `promotes` and `add_subsystem` (when inputs are promoted), which specifies the apparent shape of the promoted input (if connecting to it) or the related AutoIVC output (if left unconnected).
```
JustinSGray, line 36: I don't see how we can make it work as an argument to `add_subsystem`. In the other cases (`set_input_defaults`, and `promotes` you have the granularity to break out specific inputs or subsets of inputs that share common shape and src_indices. In `add_subsystem` you don't. 

We don't currently allow src_inidices in `add_subsystem` and I don't think we should allow `src_shape` there either. I think that using the promotes arguments for `add_subsystem` should just be a slightly less capable API

JustinSGray (changes requested):
We need to add some examples that use set_input_defaults. We also need to call out specific error cases and give a specific format for reporting those errors that will give enough clarity to the user

POEM_035.md
robfalck: I think so but I also think there's no way to know the shape of the relevant input without tihs.  Rather than attempting to deduce this ourselves, I think it's best to require the user to be explicit about this.

POEM_035.md
robfalck: Allow `src_shape` to be a dict.

```
self.add_subsystem('foo', promotes_inputs=['a', ('b', 'b_in'), 'c'],
                                    src_indices={'b_in': src_idxs},
                                    src_shape={'b_in': (4, 3))
```

POEM_035/case3.py
robfalck: Yeah I think that's better

PR #79, opened 2020-11-01
Added POEM 036: Serialization of Kriging training weights

Dakror:


Dakror:
I'm currently working on an implementation PR, so more details or changes might arise even from my side as im working on the implementation.

Dakror:
While the suggested format is JSON, in the production case I'm using this change i discovered that the saving time of the JSON is immense for large enough datasets. We are training a net with multiple thousand points and the resulting json file is gigantic. I'm therefore now adding / changing it to HDF support, at least on the local side. Feel free to comment on this.

JustinSGray:
HDF support would require a dependency that I don't think OpenMDAO currently has. It could be made into an optional dependency, but I would rather avoid it if we could. 

Depending on how you save the data, it can be shrunk substantially. What kind of data are you saving? is it just numerical? can you use numpy's binary output formats? 

Have you tried zipping the data before you save it to json: https://medium.com/@busybus/zipjson-3ed15f8ea85d

Dakror:
Yeah HDF support is currently not given thus my initial goal to use the given possibilities. I'm just saving the several NumPy arrays the Kriging model creates. I'm gonna try and look into the Numpy File format. Putting zipped B64 data into JSON seems awfully hacky to me.

Dakror:
Numpy Support using the NPZ file format was trivial to integrate and both performance and file size are excellent. Will adopt the proposal

robfalck (approved):

PR #82, opened 2020-12-02
POEM 039: Function Wrapping Component (FuncComp)

robfalck:


JustinSGray:
after discussion by the dev team, this concept was discarded in favor or a different one that relied on ExecComp instead. The new concept was given POEM_039 instead, since no significant public discussion was made on this concept.

PR #98, opened 2021-02-19
POEM 043: No src_indices warning when both components are distributed

markleader:


anilyil:
Hi @markleader. We are also interested in this modification for the MPhys package. In fact, I think we also want to disable the warnings when we connect stuff from a serial component to a distributed component; however, there is still some value in keeping that warnings for some users. 

@JustinSGray, how about we add an optional argument to add_output, say sth like `pass_shape_to_input` or `no_warning_indices`, when set to true, would stop printing the warnings. This keeps the warnings to newcomers when they are not sure about the code behavior and printing of the assumed source indices are useful, vs we can just disable these when we know the assumed indices work correctly.

JustinSGray:
@anilyil I hadn't considered per output control over the warning. Its an interesting idea, but im not convinced is the right global solution. 

there was a broader discussion of setting up a broader scheme for user control over the myriad warnings that OpenMDAO can generate. We could also consider rolling these warnings into that rework instead. 

Would it be better to just be able to turn these warnings off globally? Turning them off for specific outputs seems both tedious, and still somewhat error prone.

robfalck:
I'm still learning towards the global settings/options approach for things like this.

anilyil:
@robfalck do you have a time frame in mind to implement that? We are also okay with that approach.

robfalck:
I think I can put together a poem and have an implementation in a couple of weeks.  I don't think it's a fundamentally complex task, but getting it right will take some thought.

anilyil:
Sounds good to me. I am interested in the time frame because we are getting quite a bit of warnings in MPhys and wanted to clear those out when we are fixing the src_indices stuff. If this will warning stuff will be addressed on the OpenMDAO side, then I think we can make the modifications required to get rid of the deprecation warnings by modifying the MPhys code and call it a day.

JustinSGray:
Im going to merge this, but this POEM is going to be rejected in favor of POEM_044 which is more general and encompassing.

PR #99, opened 2021-02-25
POEM 041: Add expressions to ExecComp after instantiation

robfalck:


Kenneth-T-Moore (approved):
need to get this updated to alleviate confusion.

PR #111, opened 2021-06-21
POEM 051 - Modifications to relative step sizing during fd approximation.

Kenneth-T-Moore:


POEM_051.md
```diff
@@ -0,0 +1,107 @@
...
+absolute value ensures that we won't be taking a forward step when we request backward.
+
+
+Allow the user to specify a relative step for every element in the vector.
+==========================================================================
+
+This gives the user finer control over the approximation. For `declare_partials`, the simplest API
+would be to allow the user to pass a list (or tuple) with the same length as the input whose
+derivatives are being approximated. Here is a simple example for a 3-wide input 'x':
+
```
JustinSGray: this is more complex than we need. The user can provide a single relative step size. It should just be computed relative to each element of the vector.

POEM_051.md
```diff
@@ -0,0 +1,107 @@
...
+   self.declare_partials(of='*', wrt='x', method='fd',
+                         step=[1e-7, 1e-5, 1e-6], form='forward', step_calc='rel')
+```
+
+The key difference here is that relative step_size is computed individually for each vector element:
+```
+    rel_step_j = step_j * np.abs(x_j)
+```
+
+This API might be able to be expanded to allow specification of other aspects of the approximation:
```
JustinSGray: I don't think we need this level of flexibility. Just allowing for a more intuitive step size calc based on each element (instead of some aggregate of teh whole vector) is enough.

POEM_051.md
```diff
@@ -0,0 +1,98 @@
...
+    the old behavior. In the next release, "rel" will be the same as "rel_avg".
+
+In summary:
+
+| step_calc       | Computed Step                                   |
+| :---            |    :----:                                       |
+| "rel_avg"       | Average absolute value of the vector.           |
+| "rel_element"   | Absolute value of each vector element.          |
+| "rel_legacy"    | Norm of the vector.                             |
+| "rel_avg"       | "rel_legacy" during transition, then "rel_avg"  |
```
JustinSGray, line 97: do you mean `rel` for the last row here?

JustinSGray (approved):
other than the minor typo in the last row of the table at the end. this looks ok to me.

POEM_051.md
Kenneth-T-Moore: Thanks, typo fixed.

PR #120, opened 2021-08-19
POEM_55 - Min/Max display for array printing

andrewellis55:


JustinSGray:
This is a great idea. It  will certainly make the debug viewing of large arrays easier. Its also probably more useful than an array norm. 

I have a modest quibble with your proposed output format though 

you suggest: 
```
name           value                            size  lower   upper
-------------  -------------------------------  ----  ------  -----
stress.stress  [1.75192149e-06 1.00000000e+01]  100   -1e+30  10.0
```
Where you replace the value with a `[<lower>, <upper>]` type format. 
I worry that this is visually a bit ambiguous here. A casual user who hadn't read the docs might think its a length 2 array. There is no indication that its a min/max value there to them. 

I don't have a great alternative suggestion though. We could just change the header to say min/max (and for scalars you'd just get a scalar...). Alternatively, we could keep the current output formating, and add a new option to the `list_` methods like `print_min_max` and then add an extra data field similar to the existing one used for the array value. 

So I guess I don't like the sort-of-array look of your current proposal. I love the general idea though. So we either need a different look and header for the data in what is currently the `value` column, or we need to use a different field

andrewellis55:
I like the column option! And I agree that I'm not a huge fan of the formatting I originally proposed either.

If we went with columns, the user call could look almost exactly the same as before and just have min and max be part of the options list.

```python
prob.list_problem_vars(
    desvar_opts=['lower','min', 'max', 'upper'],
    cons_opts=['lower','min', 'max', 'upper'],
)
```
Since you've already written `_write_var_info_table` to be pretty modular, all that would need to be changed is to add parsing for a min and max column like so

```python
    def _write_var_info_table(self, header, col_names, meta, vals, print_arrays=False,
                              show_promoted_name=True, col_spacing=1):
        """
        Write a table of information for the problem variable in meta and vals.

        Parameters
        ----------
        header : str
            The header line for the table.
        col_names : list of str
            List of column labels.
        meta : OrderedDict
            Dictionary of metadata for each problem variable.
        vals : OrderedDict
            Dictionary of values for each problem variable.
        print_arrays : bool, optional
            When False, in the columnar display, just display norm of any ndarrays with size > 1.
            The norm is surrounded by vertical bars to indicate that it is a norm.
            When True, also display full values of the ndarray below the row. Format is affected
            by the values set with numpy.set_printoptions
            Default is False.
        show_promoted_name : bool
            If True, then show the promoted names of the variables.
        col_spacing : int
            Number of spaces between columns in the table.
        """
        abs2prom = self.model._var_abs2prom

        # Get the values for all the elements in the tables
        rows = []
        for name, meta in meta.items():

            row = {}
            for col_name in col_names:
                if col_name == 'name':
                    if show_promoted_name:
                        row[col_name] = name
                    else:
                        if name in abs2prom['input']:
                            row[col_name] = abs2prom['input'][name]
                        elif name in abs2prom['output']:
                            row[col_name] = abs2prom['output'][name]
                        else:
                            # Promoted auto_ivc name. Keep it promoted
                            row[col_name] = name

                elif col_name == 'value':
                    row[col_name] = vals[name]
                elif col_name == 'min':              # Added
                    row[col_name] = min(vals[name])  # Added
                elif col_name == 'max':              # Added
                    row[col_name] = max(vals[name])  # Added
                else:
                    row[col_name] = meta[col_name]
            rows.append(row)
...
```

This would then print the following from my example script (with the min and max option added to both the constraint and the desvars). (Note that the order of the columns is up to the user based on the order of `cons_opts=['lower','min', 'max', 'upper']`)

```
----------------
Design Variables
----------------
name    value         size  lower  min                max                upper  
------  ------------  ----  -----  -----------------  -----------------  ----- 
radius  [5.53128897]  1     0.0    5.531288965510812  5.531288965510812  1e+30  

-----------
Constraints
-----------
name           value                size  lower   min                     max               upper  
-------------  -------------------  ----  ------  ----------------------  ----------------  ----- 
stress.stress  |37.56105229124744|  100   -1e+30  1.7519214920178306e-06  9.99999999999987  10.0   

----------
Objectives
----------
name         value          size
-----------  -------------  ----
stress.area  [96.11744123]  1
```

For arrays it would print the min and max and for scalers it would just print the value twice.

My only remaining question would be if we should do anything about rounding. With the numpy arrays that the values are in, there are some default rounding options that apply to printing, however once the `min`/`max` is called bringing the values out into a float, we get likely a lot more decimal places than needed.

andrewellis55:
Additionally, pyOptSparse has a column of it's printout called 'status' that provides a quick visual indication if a value is at or exceeded it's upper, lower or lower bound (u/l for being at the lower/upper bound. U/L for exceeding the upper/lower bound).

If we're extracting the min/maxes anyway and we have the constraint data. Could this potentially be a worthwhile column to add as an option as well?

```
Optimization Problem -- Optimization using pyOpt_sparse
================================================================================
    Objective Function: _objfunc

    Solution:
--------------------------------------------------------------------------------
    Total Time:                    0.0090
       User Objective Time :       0.0020
       User Sensitivity Time :     0.0020
       Interface Time :            0.0050
       Opt Solver Time:            0.0000
    Calls to Objective Function :      13
    Calls to Sens Function :           10


   Objectives
      Index  Name                   Value          Optimum
          0  stress.area     8.593104E+01     0.000000E+00

   Variables (c - continuous, i - integer, d - discrete)
      Index  Name       Type      Lower Bound            Value      Upper Bound     Status
          0  radius_0      c     0.000000E+00     5.229983E+00     1.000000E+30

   Constraints (i - inequality, e - equality)
      Index  Name          Type          Lower           Value           Upper    Status  Lagrange Multiplier (N/A)
          0  stress.stress    i  -1.000000E+30    6.114275E-01    1.000000E+01              9.00000E+100
          1  stress.stress    i  -1.000000E+30    1.000000E+01    1.000000E+01         u    9.00000E+100
          2  stress.stress    i  -1.000000E+30    4.564281E+00    1.000000E+01              9.00000E+100

--------------------------------------------------------------------------------
```

JustinSGray:
i support the core idea of this poem. Adding an extra column, or finding the right overall formatting is the key need here. If you want to add an `u/l` marker somehow, I think that would be worthwhile too. 

You should remove the implementation details from the poem itself. its good to note that the implementation would be strait forward, but keep the POEM focused on the API itself. Put effort into fully spec-ing out the way the prints should look. 

We'll gladly accept an implementation PR along with the POEM, but they should be kept separate. That way, even if someone else ends up writing the implementation they have a well defined requirement to work toward. 

Im looking forward to what you come up with here!

andrewellis55:
POEM has been modified to remove the implementation. PR #2240 has been created to add in the min/max columns. I also added a few lines to make the rounding consistent between the numpy arrays and the floats.

As for the status column, I took a quick look into it, but it would depend on the driver tolerance to check if we're actually violating a bound. This would be a pretty easy thing for SLSQP, but I'm a little less familiar with the rest of the drivers that can be used so I think I'll leave that out for now and perhaps it can be added in an additional PEOM/PR/Issue in the future

robfalck (approved):

PR #121, opened 2021-09-17
Poem 056 - function metadata API

naylor-b:


POEM_056.md
```diff
@@ -0,0 +1,349 @@
...
+assume that is has the value 1.0.  If the `shape` is provided and either the default value is
+not provided or is provided as a scalar value, then the assumed default value will be
+`np.ones(shape) * scalar_default_value`, where `scalar_default_value` is 1.0 if not specified.
+If `shape` is provided along with a non-scalar default value that has a different shape, then
+an exception will be raised.
+
+
+## Setting non-default function metadata
+
+The `metadata` decorator can be used to specify metadata that is intended to apply to the function
```
JustinSGray, line 221: `decorator` isn't actually used any more. Replace with `metadata method`

POEM_056.md
```diff
@@ -0,0 +1,349 @@
...
+```python
+import openmdao.func_api as omf
+```
+
+A two-way API as described above, where API functions are used to both attach and retrieve data,
+allows the underlying data layout and location to be hidden and potentially updated later without 
+negatively impacting users of the API.
+
+This API consists primarily of methods on the OMWrappedFunc class, which wraps a plain python or
+potentially annotated function and stores metadata needed by openmdao along with that function.
```
JustinSGray, line 53: You say "potentially annotated" but don't describe what you mean by that. 

Are you saying that you could accept the I/O annotations laid out in POEM 052? If so, this POEM should stand alone, so please pull the relevant details into here and give at least one example. 

Im not sure what the annotations really buy you here though. If we track the order of the `add_input` and `add_output` calls then you wouldn't even need to provide the arg names (at least for inputs and cases where the output names could be inspected). 

I think the annotations are not needed, and we'd be better of giving examples of how we could use the AST to iner the names instead (and explain the case when return values are expressions when inference isn't possible and names will be required).

POEM_056.md
```diff
@@ -0,0 +1,349 @@
...
+
+# names of return values are 'd' and 'e' and they have no other metadata
+f = omf.wrap(func).add_outputs(d={}, e={})  
+```
+
+As mentioned above, if the function's return values are simple variable names, we don't need to
+call `output_names` because we can determine the names from inspecting the function, e.g., 
+
+
+```python
```
JustinSGray, line 271: It's a little unclear here if you're implying that you could skip the `add_input` and `add_output` calls all together here and just wrap the function. I think thats what you mean, but you should expand the example to clarify. 

can you call `add_input` without giving a name? will it follow the argument order in that case?

POEM_056.md
```diff
@@ -0,0 +1,349 @@
...
+
+f = (omf.wrap(func)
+        .declare_partials(of='*', wrt='*', method='cs')
+        .declare_coloring(wrt='*', method='cs')
+        .defaults(shape=4))
+```
+
+The args for the `declare_partials` and `declare_coloring` methods match those
+of the  `declare_partials` and `declare_coloring` methods of an OpenMDAO component.  Multiple calls
+can be made to either decorator to set up different partials/colorings.
```
JustinSGray, line 330: is that true in regular om with respect to coloring? Can you color some stuff with CS and others with FD? I thought we made it so you had to pick one or the other. 

I know its true for partials. Just not sure about coloring.

POEM_056.md
```diff
@@ -0,0 +1,349 @@
...
+
+Note that for a regular OpenMDAO component, the `method` argument can have values of 'fd' or 'cs'.
+This function based API allows one additional allowed value, 'jax', which specifies that the 
+'jax' library be used to compute derivatives using AD.
```
JustinSGray, line 334: need to clarify a bit why `jax` is not being supported for regular OM components. Its because the functional API matches what JAX expects and doesn't require us to create custom types that we pass through that somehow map to our arrays... 

I think its possible to get jax to work for regular OM components, but thats a different poem :)

JustinSGray (changes requested):
I like this better than my own POEM 052. I agree the functional interface is cleaner. I also like the fancy stacked calls :) 

I do have a few minor comments. Mostly clarifications, but I think you need to either work out the annotations or drop them completely. My vote is to drop them though. I don't think they are necessary, given what you have here.

POEM_056.md
naylor-b: You're right.  Looks like we don't support multiple calls to declare_coloring for a given component.

POEM_056.md
naylor-b: We could definitely allow `add_input` to be called without providing a variable name since we can get the name by inspecting the function, but I'm not sure we want to.  It just seems like we'd be making it easier for a user to make a mistake.  We could potentially allow the same thing in the case of outputs if their name can be inferred from the source code, but I hesitate to allow that for the same reason.

POEM_056.md
naylor-b: I updated this to say that 'jax' may be added later for regular OM components as well but is currently not supported.

POEM_056.md
naylor-b: I dropped the 'potentially annotated' stuff and added a little more explanation of output name inference.

POEM_056.md
naylor-b: all of the occurrences of 'decorator' should be gone now.

POEM_056.md
naylor-b: I cleared up the explanation of this.

robfalck (approved):

PR #128, opened 2021-11-15
POEM 059 - Unitless and Percentages

andrewellis55:


robfalck:
I can see the need for a unitless quantity that is checked for consistency when connected to others.  I think my preference would be to avoid `'none'` as a string, since it's too easily confused with `None.`  I'd suggest `'unitless'` instead.  People are unlikely to use `'unitless'` in a composite unit specification, so the length of the descriptor isn't a big deal there.

So in short...
- units of `None` will continue to operate as the currently do...a dimensionless quantity that isn't enforced during connection.
- units of 'unitless' will serve as a dimensionless quantity that is enforced during connection.

robfalck:
Regarding the `percent` and `dectopercent` units...aren't those really just dimensionless quantities?  We might go down a deep rabbit hole if we start enforcing connection of some dimensionless quantities (percentages) and not others (mach number, aspect ratio, CL/CD, etc).

Mind you, you as a user are entitled to add your own units to the unit system...I'm just not sure it's something that should be part of the standard installation.

andrewellis55:
> I can see the need for a unitless quantity that is checked for consistency when connected to others. I think my preference would be to avoid `'none'` as a string, since it's too easily confused with `None.` I'd suggest `'unitless'` instead. People are unlikely to use `'unitless'` in a composite unit specification, so the length of the descriptor isn't a big deal there.
> 
> So in short...
> 
> * units of `None` will continue to operate as the currently do...a dimensionless quantity that isn't enforced during connection.
> * units of 'unitless' will serve as a dimensionless quantity that is enforced during connection.

I agree with swapping to 'unitless' from 'none'. I've updated the POEM and I will update the PR momentarily

andrewellis55:
> Regarding the `percent` and `dectopercent` units...aren't those really just dimensionless quantities? 

In a sense yes, but they fall under the broader category of the 'parts-per' notation which is defined by Wikipedia as "a set of pseudo-units to describe small values of miscellaneous dimensionless quantities".

However, raising the point of percentages just being a scalar of unitless numbers made me realize that a much simpler implementation is possible. What I was referring to as 'decimal' is really the same as the 'unitless' quantity. This means that percent could be defined in the unit_library.ini as 

```
percent: unitless/100, percentage
```

And we could run my above example as 

```python
import openmdao.api as om

class DectoPercent(om.ExplicitComponent):
    def setup(self):
        self.add_input('val', units='unitless')

if __name__ == '__main__':

    prob = om.Problem()
    prob.model.add_subsystem('val_comp', DectoPercent(), promotes=['*'])

    prob.setup()
    prob.set_val('val', 10, 'percent')

    print('Dec: ', prob.get_val('val', 'unitless'))
    print('%: ', prob.get_val('val', 'percent'))
```
Output
```
Dec:  0.1
%:  10.0
```

A more worked example is shown bellow that shows some of the utility. When a margin of safety is calculated, it may be unlear when getting the value if it's being returned as a unitless number or as a percentage. This can be made more clear to a user with the implementation of this PEOM

```python
import openmdao.api as om

class MarginOfSafety(om.ExplicitComponent):
    def setup(self):
        self.add_input('stress', units='MPa', val=392)
        self.add_input('tensile_strength', units='MPa', val=400)

        self.add_output('margin_of_safety', units='unitless')

    def compute(self, inputs, outputs):
        stress = inputs['stress']
        tensile_strength = inputs['tensile_strength']
        outputs['margin_of_safety'] = tensile_strength / stress - 1

if __name__ == '__main__':

    prob = om.Problem()
    prob.model.add_subsystem('margin_comp', MarginOfSafety(), promotes=['*'])

    prob.setup()
    prob.run_model()

    print('Margin of Safety (%): ', prob.get_val('margin_of_safety', units='percent'))
```
Output
```
Margin of Safety (%):  [2.04081633]
```

JustinSGray:
I agree with defining percent at unitless/100

andrewellis55:
PR and POEM have been updated based on my last comment

PR #131, opened 2021-11-29
Added POEM 060 - HTML Report Generation

robfalck:
This POEM would convert some optional OpenMDAO HTML outputs to reports which are automatically created and saved in a subdirectory.  This will aid users by providing feedback that is always available, without having to figure out how to enable it.

### DIFF
--- a/POEM_060.md
+++ b/POEM_060.md
@@ -7,9 +7,9 @@ Associated implementation PR: N/A
 
 Status:
 
-- [x] Active
+- [ ] Active
 - [ ] Requesting decision
-- [ ] Accepted
+- [x] Accepted
 - [ ] Rejected
 - [ ] Integrated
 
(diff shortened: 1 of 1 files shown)


POEM_060.md
```diff
@@ -0,0 +1,62 @@
...
+All reports will be saved within a subfolder to avoid cluttering the current working directory and making cleanup easier.
+Reports will be saved as HTML documents.
+
+A good name for the subfolder might be `f'{sys.argv[0]}_{problem_name}_reports'` in the current working directory, so that multiple run files in the CWD don't overwrite reports made by each other.
+Add `report_dir` as an option to Problem so that the user can override the default.
+
+### Disabling report generation
+
+Experienced users may not want to deal with generating reports on each run of their model, especially if their systems are especially large and the report generation is too time consuming.
+This can be done by using a `reports` argument to Problem.setup (similar to checks), or by setting an environment variable `OPENMDAO_REPORTS`, where a value that is one of `['0', 'false', 'off']` (noisily) disables all report generation.
```
swryan, line 37: why not an option?   alongside the `report_dir` option above...

robfalck: Agreed, an option is probably the right place for this.

PR #144, opened 2022-07-23
POEM 068

lamkina:
System output caching for nonlinear solvers to enable a `restart` mode for model evaluations within an optimization.

### DIFF
--- a/POEM_068.md
+++ b/POEM_068.md
@@ -1,5 +1,5 @@
 POEM ID: 068  
-Title:  Nonlinear Solver State Caching  
+Title:  Nonlinear Solver System Output Caching  
 authors: @lamkina  
 Competing POEMs:  
 Related POEMs:  
--- a/POEM_068.md
+++ b/POEM_068.md
@@ -44,7 +44,158 @@ the cached states become an increasingly better initial guess for the nonlinear 
 
 ## Description
 
-This POEM implements the *restart* method using cached states within the nonlinear solver.
+### Summary
+
+This POEM implements the *restart* method using system output caching in the nonlinear solver base class. A
+reference implementation can be found [here](https://github.com/lamkina/OpenMDAO/blob/solver_cache/openmdao/solvers/solver.py).
+
+The premise behind the method is as follows:
+
+1. After a successful model evaluation involving a nonlinear solver, cache the system output vector.
+2. If the model fails to converge in a subsequent iteration, set a *previous evaluation fail flag* to true.
+3. Upon the next model evaluation, replace the current system outputs with the cached outputs before running
+   the nonlinear solver.
+4. Evaluate the model with the cached outputs as the initial guess for the nonlinear system.
+
+The caching feature will exist within the `NonlinearSolver` base class and work synergistically with the MAUD framework by only caching states at or below the current system.  Additionally, users will only see one change to the nonlinear solver interface, making this feature easy to learn and add to new and existing models.
+
+### Changes to the Nonlinear Solver Interface
+
+The addition of system output caching introduces a change to user interaction with the nonlinear solver interface.
+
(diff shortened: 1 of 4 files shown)


robfalck:
Thank you for the extremely thorough proposal!  This seems like it will be doable.  The dev team will pass this around and assuming there are no unforeseen issues we should be able to get a reference implementation integrated.

POEM_068.md
```diff
@@ -0,0 +1,198 @@
...
+
+
+
+## Motivation
+
+Many optimizers employ backtracking algorithms to recover from failed model evaluations.
+After a failed evaluation, the user in conjunction with OpenMDAO must decide how to initialize the model's states
+for the next backtracking iteration.  Three main options exist:
+
+1. **Cold Start:**  Use the initial guesses to the model at the onset of the optimization problem.
```
nwu63, line 24: I just had a quick chat with @lamkina about this. A few things that I brought up:
- In the optimization world, personally I consider "warm start" a different approach to what's described here. Cold vs. warm vs. hot refers to the _quantity_ of information supplied to initialize some model, not the _quality_, so the warm start described here would actually be a cold start with a different set of states. That being said, I'm not sure if the same translates to OM and there's little reason to change established terminology.
- I would prefer a different name to "restart" though, since it's not very descriptive and warm start can be described as a restart strategy also. @lamkina suggested something along the lines of the following:
  - `warm start` -> `restart from previous`
  - `restart` -> `restart from successful`
- whatever is picked here should match the name of the actual option (currently `use_cached_states`).

POEM_068.md
```diff
@@ -0,0 +1,198 @@
...
+- [ ] Requesting decision
+- [ ] Accepted
+- [ ] Rejected
+- [ ] Integrated
+
+
+
+## Motivation
+
+Many optimizers employ backtracking algorithms to recover from failed model evaluations.
```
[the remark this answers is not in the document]
lamkina: Yes this is about restart methods for solvers.  Including the optimizer in the motivation serves as the prime example for where this solver restart method will have the most value.

POEM_068.md
lamkina: The terminology should be changed to maintain consistency with @nwu63 definition.  We should be describing the quantity or type of inormation being supplied as well as from where the information originates.

naylor-b:
@lamkina I have an implementation branch here:  https://github.com/naylor-b/OpenMDAO/tree/solver_output_cache if you want to try it out.  The discussion above mentions renaming `use_cached_states`.  The current implementation still uses that name.  If there's a new name you prefer, let me know and I'll update it.

lamkina:
@naylor-b I took a brief look and I will try it out later today.  @nwu63 and I think that the following naming options make sense:
1. For `warm starts` (the current OM implementation) we would use the term `restart from previous`
2. For `restarts` (My proposed idea) we would say `restart from successful`

This difference makes it clear to the user what is happening in each case.  Since we will want to turn this on/off with a single option, I think it makes sense to change the options from `use_cached_states` to `restart_from_successful`.  Then in the options documentation we can explain the nomenclature difference between when that option is on/off.

Does this make sense or do you have a better naming scheme?

naylor-b:
Changing `use_cached_states` to `restart_from_successful` seems reasonable to me, so unless someone has an objection, I'll go ahead and make that change.

robfalck:
@lamkina Have you had a chance to test Bret's PR [here](https://github.com/OpenMDAO/OpenMDAO/pull/2590)?

lamkina:
@robfalck yes I have tested Bret's PR with my test problems and my aeropropulsive work with MPhys.  It worked great so I moved the POEM status to `requesting decision`.

lamkina:
I also updated the POEM to reflect the new naming scheme.  Bret is using the new naming scheme in his PR as well.

PR #145, opened 2022-08-04
POEM 069: Declare residual names for implicit components

joanibal:
POEM to address confusion in residual naming. Coauthored with @eytanadler.

### DIFF
--- a/POEM_069.md
+++ b/POEM_069.md
@@ -1,5 +1,5 @@
-POEM ID: 069
-Title: Declare residual names for implicit components
+POEM ID: 069  
+Title: Declare residual names for implicit components  
 authors: @joanibal, @eytanadler  
 Competing POEMs:  
 Related POEMs:  
--- a/POEM_069.md
+++ b/POEM_069.md
@@ -1,0 +1,167 @@
+POEM ID: 069
+Title: Declare residual names for implicit components
+authors: @joanibal, @eytanadler  
+Competing POEMs:  
+Related POEMs:  
+Associated implementation PR: [#2589](https://github.com/OpenMDAO/OpenMDAO/pull/2589)
+
+Status:
+
+- [x] Active
+- [ ] Requesting decision
+- [ ] Accepted
+- [ ] Rejected
+- [ ] Integrated
+
+
+## Motivation
+
+In many cases, it does not make sense to name the residuals of an implicit component by its outputs.
+Furthermore, the residual may not even depend on the state by which it is named.
+Currently, the keys of the residual vector are automatically set to match the keys of the output vectors. 
+
+This can cause confusion when working with implicit components.
+Consider the following example from the [circuit analysis example](https://openmdao.org/newdocs/versions/latest/examples/circuit_analysis_examples.html):
+
+```python
+class Node(om.ImplicitComponent):
+    """Computes voltage residual across a node based on incoming and outgoing current."""
+
+    def initialize(self):
+        self.options.declare('n_in', default=1, types=int, desc='number of connections with + assumed in')
+        self.options.declare('n_out', default=1, types=int, desc='number of current connections + assumed out')
+
+    def setup(self):
+        self.add_output('V', val=5., units='V')
+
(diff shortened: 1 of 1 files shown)


JustinSGray:
Thanks for the POEM! we appreciate the input here. Some of the internal NASA users immediately agreed that matching states to residuals has always felt weird to them. 

I was minorly concerned about a mathematical problem with randomly associating the order of residuals and states. However, I worked out the equations and its fine as long as the following limitations are followed: 

1) within a component you still must have the same total number of states and residuals 
2) (basically a restate of #1 but I think the clarity is needed) You can not split up states/residuals across multiple components. 

The net effect of these restrictions is that it should be possible to implement this change in a "cosmetic" manner at the component level. As far as the rest of OpenMDAO is concerned each state will still have a direct 1-to-1 mapping into the residual array.

POEM_069.md
```diff
@@ -0,0 +1,167 @@
...
+In this example, the fact that one residual is accessed with `"x"` and the other with `"y"` is arbitrary.
+There is no apparent relationship between the output names and the residuals.
+For example, `residuals["y"]` has no relation to `y`.
+Declaring and implementing the derivatives can be even more confusing because the keys don't imply a relationship with the residuals.
+A new user would look at `partials["x", "x"]` and wonder why it isn't equal to 1 (speaking from experience...).
+
+
+## Proposed solutions
+
+We propose adding the implicit component method `add_residuals` to allow the user to explicitly declare the residuals.
```
naylor-b, line 94: should be `add_residual` instead of `add_residuals`

POEM_069.md
```diff
@@ -0,0 +1,167 @@
...
+            i_name = 'I_in:{}'.format(i)
+            self.add_input(i_name, units='A')
+            self.declare_partials('V', i_name, val=1)
+
+        for i in range(self.options['n_out']):
+            i_name = 'I_out:{}'.format(i)
+            self.add_input(i_name, units='A')
+            self.declare_partials('V', i_name, val=-1)
+        
+        self.add_residual('I_net', val=0., desc='net current flowing through the node')
```
naylor-b, line 120: In the description of `add_residual` that comes later, it says the args are (name, shape), but here it also takes 'val' and 'desc'.  The POEM should include the full list of arguments that can be passed to `add_residual`.  Also, the existing `add_output` method supports `res_ref` for residual scaling and `res_units`.  Will the `add_residual` method have corresponding `ref` and `units` args?

naylor-b:
OpenMDAO currently supports the writing of residual values to its data recorders, and filtering those by name.  This seems to imply that we would need to propagate these new residual names up the system tree which would require adding additional data structures outside of just the components to support this.

POEM_069.md
eytanadler: Agreed. Would `val` get used anywhere if we did specify it? As far as I know, it wouldn't be used by a monolithic solver, but maybe for another case?

naylor-b: I can't think of any cases where `val` would actually be used.  We probably just shouldn't allow `val` as an arg.

eytanadler:
> OpenMDAO currently supports the writing of residual values to its data recorders, and filtering those by name. This seems to imply that we would need to propagate these new residual names up the system tree which would require adding additional data structures outside of just the components to support this.

@joanibal and I are not sure what this would involve. Is it a showstopper for adding the `add_residual` method?

POEM_069.md
eytanadler: Agreed. Would `val` be used by the solver somehow if specified? From my understanding of the monolithic solver it would not, but I'm not sure if it may be different for other cases.

eytanadler: Ok, sounds good. We'll make the change to the POEM. (sorry about the duplicate comment, my computer was struggling)

naylor-b:
It's not a showstopper, just an indication that this might require a little more than just introducing a mapping at the component level.

robfalck (approved):

PR #147, opened 2022-09-02
POEM 070: Inputs report

robfalck:


Kenneth-T-Moore:
Is your suggestion for a new function, or for the current list_inputs to be modified to produce this info?  I know the POEM suggests a new report, but I find the functions useful when live debugging. I think implementation could be just an extra argument to filter down to the inputs that are only connected to IVCs. The extra column(s) could be added here too.

Otherwise, I think the IVC connection column isn't useful, since every input is connected to an IVC (unless you want to inform whether a manually created IVC is present.)  I may be a little confused whether this is a generic "inputs" report for all inputs in the model, or if the report is just for unconnected inputs, as in the motivation. If the former, then disregard.

robfalck:
An input connected to the output of some other component is, in general, not reliant on the user to provide a good starting value for it.

Any input that's "unconnected" (AutoIVC) or explicitly connected to an IVC in general is an input that the user at least needs to be aware of.

Regarding if we produce this table with list_inputs, I agree that it would be useful for live debugging and I suspect the implementation of this would probably involve adding something like an `output_fmt='html'` option to list inputs.
But listing inputs is one of those things that's cheap and the user should have access to just by virtue of executing a model.

PR #148, opened 2022-09-02
POEM 071: Change ExecComp to use `declare_coloring`

robfalck:


POEM_071.md
```diff
@@ -0,0 +1,41 @@
...
+
+
+## Motivation
+
+The `ExecComp` in OpenMDAO uses complex-safe functions to provide user-defined calculations within a run script in leiu of writing a custom component.
+
+The option `has_diag_partials` on ExecComp was added prior to the development of partial derivative coloring to provide a common sparsity pattern for vectorized components.
+
+With the advent of partial coloring, `has_diag_partials` is no longer needed.
+Users should be able to expect that `ExecComp` will determine the sparsity pattern of calualtions involved.
```
naylor-b, line 24: spelling:  calualtions

Kenneth-T-Moore (approved):

POEM_071.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Changes
+
+1. `has_diag_partials` will be marked as deprecated but still be allowed as an option that does nothing.
```
naylor-b, line 28: We ended up leaving the old internal CS stuff intact because that allows us to use CS local to the ExecComp without requiring the full model vectors to be allocated as complex, and has_diag_partials offers a performance improvement over computing the coloring so I left it in (with no deprecation).

POEM_071.md
```diff
@@ -0,0 +1,41 @@
...
+2. `declare_coloring_kwargs` will allow the user to specify any arguments to be passed to [System.declare_coloring](https://openmdao.org/newdocs/versions/latest/features/experimental/approx_coloring.html#dynamic-coloring))
```
naylor-b, line 29: `declare_coloring_kwargs` wasn't used.  The thought process was that if a user has specific args they need to pass (usually not the case), then they can just call `declare_coloring` and pass them normally.  Otherwise, if they don't call `declare_coloring` explicitly, they get the internal coloring method that only uses default args.

PR #149, opened 2022-09-04
POEM 072: Ability to modify bounds and scaling after output/desvar/re…

robfalck:
…sponse creation

### DIFF
--- a/POEM_072.md
+++ b/POEM_072.md
@@ -64,7 +64,7 @@ prob.model.set_constraint_options('alias' or 'path.to.con', equals=_unspecified,
 prob.model.set_objective_options('path.to.obj', scaler=_unspecified, adder=_unspecified, ref=_unspecified, ref0=_unspecified)
 ```
 
-One complication is the path to constraints in `set_constraint_options`, since a single variable may have multiple constraints imposed upon it provided they are aliased, but we should be able to account for this.
+One complication is the path to constraints in `set_constraint_options`, since a single variable may have multiple constraints imposed upon it provided they are aliased, but we should be able to account for this.  If there is ambiguity (an alias exists), raise an error.
 
 Note that we can arrive at a situation in which, if we're redefining `scaler/adder/ref0/ref`, where all 4 are specified. If the user overrides any of these, the others should be reset to `None`.
 
--- a/POEM_072.md
+++ b/POEM_072.md
@@ -6,6 +6,7 @@ Competing POEMs:
 Related POEMs:  
 Associated implementation PR: N/A  
 
+
 Status:
 
 - [x] Active
--- a/POEM_072.md
+++ b/POEM_072.md
@@ -1,4 +1,3 @@
-
 POEM ID: 072  
 Title: Add ability to modify bounds and scaling of implicit outputs and optimizer variables after creation.  
 authors: @robfalck  
--- a/POEM_072.md
+++ b/POEM_072.md
@@ -1,0 +1,72 @@
+
+POEM ID: 072  
+Title: Add ability to modify bounds and scaling of implicit outputs and optimizer variables after creation.  
+authors: @robfalck  
+Competing POEMs:  
+Related POEMs:  
+Associated implementation PR: N/A  
+
+Status:
+
+- [x] Active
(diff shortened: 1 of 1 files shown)


POEM_072.md
```diff
@@ -0,0 +1,72 @@
...
+- [x] Active
+- [ ] Requesting decision
+- [ ] Accepted
+- [ ] Rejected
+- [ ] Integrated
+
+
+## Motivation
+
+In OpenMDAO, implicit outputs can have bounds attach to them that are enforced by bounds-enforcing linesearches within nonlinear solvers.
```
JustinSGray, line 20: Any output can have bounds. Not just implicit

POEM_072.md
```diff
@@ -0,0 +1,72 @@
...
+
+We should limit the user to changing properties associated with the nonlinear solver: `lower`, `upper`, `ref`, `ref0`, `res_ref`.  I'm currently undecided if we need to give the user the ability to change tags.
+
+When used, OpenMDAO should check, at a minimum, that the given path is a valid output.
+
+## Proposed API for Outputs
+
+```python
+
+prob.model.set_output_solver_options(name='path.to.output', lower=_unspecified, upper=_unspecified, ref=_unspecified, ref0=_unspecified, res_ref=_unspecified)
```
JustinSGray, line 45: That api name would not work with tags, which are not related to solvers

JustinSGray: Why make this a method on subsystem? Would u expect the user to grab some lower level group or comp? The path.to.var seems to imply that it should be problem level maybe

JustinSGray (commented):
What if multiple calls to the over ride methods are made in the same script (or some series of scripts). Is that an error? Last one takes precedence? Should we warn?

POEM_072.md
robfalck: If it's on system the user can use whatever depth they choose.  Doesn't seem to be a good reason to limit them by making it problem level.

POEM_072.md
robfalck: True, will correct

robfalck:
Later uses override earlier overs, don't see a need to warn users.

POEM_072.md
robfalck: Yeah I wrestled with the name on this one, for exactly that reason.  We're not allowing the user to override all options, just those pertaining to it's use with solvers.

naylor-b: `set_output_solver_options` makes sense to me.

Kenneth-T-Moore: I guess it makes sense to me too. I think I initially transposed it in my mind as set_solver_output_options, which is not very good.

POEM_072.md
```diff
@@ -0,0 +1,72 @@
...
+OpenMDAO's `add_design_var`, `add_constraint`, and `add_objective` imply that a new desvar, constraint, objective is being created.
+The new API should allow properties of those already created to be modified in place:
+
+```python
+prob.model.set_design_var_options('path.to.desvar', equals=_unspecified, lower=_unspecified, upper=_unspecified, scaler=_unspecified, adder=_unspecified, ref=_unspecified, ref0=_unspecified)
+prob.model.set_constraint_options('alias' or 'path.to.con', equals=_unspecified, lower=_unspecified, upper=_unspecified, scaler=_unspecified, adder=_unspecified, ref=_unspecified, ref0=_unspecified)
+prob.model.set_objective_options('path.to.obj', scaler=_unspecified, adder=_unspecified, ref=_unspecified, ref0=_unspecified)
+```
+
+One complication is the path to constraints in `set_constraint_options`, since a single variable may have multiple constraints imposed upon it provided they are aliased, but we should be able to account for this.
```
Kenneth-T-Moore, line 67: Would this allow you to completely redefine a constraint, like turn and equality into an inequality? (and is that intended?)

robfalck: We're venturing into big kid rules a bit here, allowing a user to mess things up.  It's a rare use case but in my past I've certainly experimented with changing equality constraints into reasonably tight lower/upper bounds to see how it changed performance.  So I suppose I can see a use for it.

PR #152, opened 2022-10-22
POEM 074: Suggest variables close in name on failed connection attempt 

joanibal:
This poem is a quality of life improvement to make failed connections easier to debug.


robfalck:
Seems like we should be able to make this happen pretty quickly.  Good suggestion.

PR #75, opened 2020-09-18
POEM_032 - driver scaling report

JustinSGray:

PR #154, opened 2022-11-28
Poem #75: Convention for distributed/serial variables and when to allreduce

kejacobson:


kejacobson:
@JustinSGray , @Kenneth-T-Moore , @anilyil , @friedenhe, @bernardopacini, @timryanb, here's what I have based on our discussion today.

POEM_075.md
```diff
@@ -0,0 +1,202 @@
...
+Conceptually, a user of a physics solver never cares about a rank's contribution to an angle of attack sensitivity or other serial variable, just the net sensitivity.
+Same with a structural solver and derivatives with respect to panel thickness or geometry parameters.
+
+2. Intermediate reverse mode states (adjoint vectors) are independent of the number of processors used with this convention.
+This is helpful because these intermediate states are used for visualization such as [figure 10 of this paper](https://arc.aiaa.org/doi/full/10.2514/1.C032189).
+
+3. When debugging, printing derivatives such as those wrt serial variables will be consistent across ranks.
+
+### "Fixing" distributed variable derivatives in OpenMDAO for the new convention
+For the final step of computing a derivative, OpenMDAO is doing an allreduce and division by number of processors. This should still be required for ParallelGroup cases and serial variables, but is the `wrt` variable is distributed, these steps should be skipped. I.e., OpenMDAO should add something like `if not variable.distributed` around this final allreduce+division step.
```
JustinSGray, line 197: typo: but *if* the 'wrt'

JustinSGray:
I like this POEM because it makes the convention easy to explain. "Serial values and derivatives should be the same everywhere" and it allows us to add that rule checking into both check_totals and check_partials. 

I think the only minor clarification that is needed here it to state that OM would be adding those checks into the two `check` methods and hence this is potentially backwards incompatible for some users.

POEM_075.md
kejacobson: fixed

PR #155, opened 2022-12-01
Poem 076, directional check totals

kejacobson:


Kenneth-T-Moore (approved):

robfalck (approved):
Approved, but noted minior grammar error to fix when we mark this as accepted.

PR #167, opened 2022-12-20
POEM 079: Raise on an initial design var value that exceeds the specified bounds.

robfalck:


JustinSGray:
I think that it would be fine to change the default behavior here. It very likely will impact users, so you *should* start with a warning. I would issue a strong deprecation warning for 2 or 3 releases. But I suggest that the POEM should commit to making this a hard error. 

I don't see a problem with enforcing dv bounds on the initial condition. Its really more correct. The only problem is the transition pain it causes users, but that can be mitigated with deprecation warnings and enough runway to let them fix it. 

The deprecation warning should say "this will become an error in version XXY" so there is no ambiguity.

robfalck:
There are potential issues with things like grid refinement in dymos where the interpolation may put the design variables outside of its bounds.  We might want to support clipping as an option for situations like that.

I have no problem with making this an error by default, I just think users should be able to override the behavior if they like.

PR #172, opened 2023-02-15
POEM 081: SubproblemComp

nsteffen:
Proposal to add a subproblem component to core components.


POEM_081.md
```diff
@@ -0,0 +1,421 @@
...
+
+- [x] Active
+- [ ] Requesting decision
+- [ ] Accepted 
+- [ ] Rejected
+- [ ] Integrated
+
+## Motivation
+
+Sometimes, the user might wish to set up a problem that has models that evalutate OpenMDAO systems themselves. However, there is currently no simple way to do this within OpenMDAO.
```
naylor-b, line 18: typo: evalutate

POEM_081.md
```diff
@@ -0,0 +1,421 @@
...
+        you wish to refer to it by within the subproblem [e.g. (prom_name, var_name)].
+    outputs : list of str or tuple
+        List of desired outputs from subproblem. If an element is a str, then it should be
+        the var name in its promoted name. If it is a tuple, then the first element 
+        should be the promoted name, and the second element should be the var name
+        you wish to refer to it by within the subproblem [e.g. (prom_name, var_name)].
+    **kwargs : named args
+        All remaining named args that become options for `SubproblemComp`.
+    """
+    def __init__(self, model=None, driver=None, comm=None, name=None,
```
naylor-b, line 77: I'm curious why this component takes a model and some Problem args, then constructs an internal problem (twice) instead of just passing a Problem instance and the list of inputs and outputs into it.  I could be missing something, but it seems like that would be a much simpler API.

naylor-b (changes requested):

POEM_081.md
nsteffen: You're right. My initial thought was to take away the steps of instantiating a problem from the user, but looking at what I have, it would definitely be easier to instantiate a problem outside of subproblem and pass that in. Also, I guess I didn't catch in the moment that the two instances of problem wasn't the most effective way of doing this, so thanks for pointing that out.
I want model to be passed as an argument along with problem too, however. This way, that extra step of adding the model as a subsystem to the problem model is handled and there is no confusion with that.
Thanks for looking through this, I will update the POEM soon.

robfalck (changes requested):

POEM_081.md
```diff
@@ -0,0 +1,404 @@
...
+        # store inputs and outputs in options
+        self.options.declare('inputs', {}, types=dict,
+                             desc='Subproblem Component inputs')
+        self.options.declare('outputs', {}, types=dict,
+                             desc='Subproblem Component outputs')
+
+        # set other variables necessary for subproblem
+        self._prev_complex_step = False
+
+        p = self._subprob = om.Problem(driver=driver, comm=comm, name=name,
```
naylor-b, line 105: I don't think we should be instantiating the problem and setting it up here in the component `__init__`.  We should just store the args and do the actual instantiation and setup in the component's setup method.  That will ensure that we can correctly set the comm.

POEM_081.md
```diff
@@ -0,0 +1,404 @@
...
+            for ip in self._input_names:
+                self.declare_partials(of=var, wrt=ip)
+    
+    def compute(self, inputs, outputs):
+        p = self._subprob
+
+        # switch subproblem to use complex IO if in complex step mode
+        if self.under_complex_step != self._prev_complex_step:
+            if self.under_complex_step:
+                p.setup(force_alloc_complex=True)
```
naylor-b, line 254: Allocation of complex vectors due to `force_alloc_complex` should only ever be done once, i.e., if we ever have to do it, we should just leave it that way from that point on instead of constantly allocating and deallocating the problem vectors.  We will still need to call problem.set_complex_step_mode(True/False) when it changes, but we don't have to keep doing the vector reallocation.

naylor-b (changes requested):

POEM_081.md
nsteffen: One of the issues I have with instantiating the problem in `setup` is that I need to call `prob.final_setup` in order to access the model's input and output lists and get the meta data for the inputs and outputs. If I move the problem instantiation to `setup`, I'll have to call for the I/O meta data every time `setup` is called, which I figured is something we want to avoid. Is there no way we can ensure the comm is set correctly in `__init__`?

POEM_081.md
```diff
@@ -100,124 +206,40 @@ class SubproblemComp(om.ExplicitComponent):
...
-                inp_dict = {inp: meta for _,meta in model_inputs.items()
-                            if meta['prom_name'].endswith(inp)}
-
-                # check if provided variable appears more than once in model
-                if len(inp_dict) > 1:
-                    raise Exception(f'Ambiguous variable {inp} in inputs. To'
-                                    ' specify which one is desired, use a tuple'
-                                    ' with the promoted name and variable name'
-                                    ' instead [e.g. (prom_name, var_name)].')
+        if self.first_setup:
```
naylor-b, line 227: I'm not sure it's wise to skip the input/output checking after the first setup call, especially since setup won't be called that often anyway.  Are we 100% sure that the model we contain hasn't been modified in some way since the last setup call that could make the checks fail?  Users do some crazy things sometimes...

robfalck (approved):

PR #176, opened 2023-04-28
POEM_083: Specifying order when adding a subsystem.

robfalck:
Proposes additional keyword arguments to `add_subsystem` that allow the user to insert it into the current subsystem list, rather than forcing it to go on the end.


robfalck:
I'm not going to approve or reject this yet, but with the recent auto ordering work this may no longer be necessary.

PR #193, opened 2024-03-06
POEM 95: Support user/developer defined callback functions

A-CGray:


naylor-b:
@A-CGray are you familiar with OpenMDAO's 'hook' system?  It's used by our report system, but it can also be used directly.  It allows you to register functions either before or after (or both) a specific method in an OpenMDAO object, for example a Problem.  So doing something like:

`_register_hook('run_model', class_name='Problem', post=my_hook_function)`

would call `my_hook_function(p)`, where p is a Problem instance, at the end of the `run_model` method.

A-CGray:
Thanks @naylor-b that does sound like it'd address the problem and driver level callbacks. Does it work at the component/system level too? i.e could the developer of a component implement a hook that will be called once at the end of `run_model` regardless of how the end user uses that component in their model? And without the user having to call `_register_hook`? If so then I think this would solve our problems

naylor-b:
The use case where the component is registering hooks on the problem or driver might be hard to do in a non-hacky way.  We may need to add something more specific for that instead of using the existing hook system.

robfalck:
I'm closing this PR as rejected.
The hook system will suffice in a lot of cases and the new post-run validate method may suffice here as well.

PR #194, opened 2024-03-15
POEM_096: Minimization of Constraint Violation

andrewellis55:


robfalck:
Thanks for initiating this.

I have a couple of questions about implementation.

1. Is the constraint violation just the L2 norm of the driver-scaled constraints? Is there any reason to support any other ways of computing it?

2. From an API standpoint, should `feasible_point=True` be an argument to `run_driver`, or an option?  I'm somewhat in favor of it being a `run_driver` argument. Since we're redefining the optimization problem after setup there may be issues with relevance
 
 This would further boost the case for refactoring the drivers into `OptimizationDrivers` and `AnalysisDrivers`, since this wouldn't make sense in the latter context.

andrewellis55:
1. I believe that SNOPT uses the L1 norm to find a feasible point 

image  
...  
image

I don't recall Dr. Martin's textbook mentioning minimization of constraint violation, but where he mentions merit functions it's suggested to use either L1 or L2  

image

I'm not an expert on the topic but perhaps some of the academic collaborators may have a suggestion\preference.

2. I agree with it being a input to run_driver. I think `feasible_point` could be a parameter name, although I could imagine some other option might be `find_feasible_point`, `feasibility_optimization` or `minimize_constraint_violation` so that the parameter name has a stronger implication that it's changing what the driver it doing (from minimizing the objective to finding a feasible point with no regard for the objective).

PR #196, opened 2024-04-26
POEM_096 Research and Pseudocode

jsrogan:
We have included an active POEM.md as well as our pseudocode of our analysis of the problem so far


POEM_096/POEM_096.md
```diff
@@ -0,0 +1,34 @@
...
+- [x] Active
+- [ ] Requesting decision
+- [ ] Accepted
+- [ ] Rejected
+- [ ] Integrated
+
+
+
+## Motivation
+Considering the fact that private optimization libraries such as SNOPT offer the ability to find a feasible starting point, it makes sense that OPENMdao should also have that functionality. Aditionally OPENMdao's current components do not calculate reverse derivatives, so the implementation of this is also necessary (and still ongoing).
```
robfalck, line 25: OpenMDAO computes partials at the component level, either through specification of a jacobian matrix - or through a "matrix-free" API (the component.compute_jacvec_prod method) in it will compute a forward JVP or a reverse VJP. The framework then assembles these calculations into a graph for efficient computation and assembly of the total derivatives. The notion of reverse derivatives just aren't relevant to the implementation of a feasible starting point.

robfalck (commented):
A feasible starting point in this case means taking the constraints, about which we already know, and forming some scalar constraint violation, and then using this as a "temporary" objective function.  In this initial optimization, we use this temporary objective, without any of the user defined constraints or objectives, and just minimize the constraint violation.  This gets us to a solution that should be feasible. From this point, we can initiate the second optimization with the full constraint vector and user-defined objective.

I'm closing this PR as it's not really hitting the mark on what we need out of a POEM 096 implementation.

PR #197, opened 2024-05-17
POEM 097: A Quantity Object to associate units with values.

robfalck:
This adds a lightweight Quantity object that associates some numerical value (any numerical value allowed by OpenMDAO...floats, or numpy arrays) with a unit.


robfalck:
The Aviary team has decided to attack this problem from a different perspective.

